PR #251, opened 2022-06-08
Refactor get matching BTC price to shared space

TonySemikin:
#### Pre-Release

- [ ] Check migrations
  - Login/logout
  - Registration
  - Update user data

#### Post-Release

- Test basic user operations

### DIFF
--- a/src/payment/models/exchange/exchange-utility.service.ts
+++ b/src/payment/models/exchange/exchange-utility.service.ts
@@ -1,0 +1,24 @@
+import { Injectable } from '@nestjs/common';
+import { BinanceService } from './binance.service';
+import { Price } from './dto/price.dto';
+import { KrakenService } from './kraken.service';
+
+@Injectable()
+export class ExchangeUtilityService {
+  constructor(private readonly krakenService: KrakenService, private readonly binanceService: BinanceService) {}
+
+  async getMatchingPrice(fromCurrency: string, toCurrency: string, matchThreshold = 0.02): Promise<Price> {
+    const krakenPrice = await this.krakenService.getPrice(fromCurrency, toCurrency);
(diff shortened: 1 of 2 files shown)


src/payment/models/exchange/exchange-utility.service.ts
```diff
@@ -0,0 +1,24 @@
...
+    const { price: krakenPriceValue } = krakenPrice;
+    const { price: binancePriceValue } = binancePrice;
+
+    if (Math.abs(binancePriceValue - krakenPriceValue) / krakenPriceValue > matchThreshold)
+      throw new Error(
+        `${fromCurrency} to ${toCurrency} price mismatch (kraken: ${krakenPriceValue}, binance: ${binancePriceValue})`,
+      );
+
+    return krakenPrice;
+  }
```
davidleomay, line 23: Should conversion only be possible for the reference assets EUR and BTC, USDT, USDC? And in case of USDC the reference price should be taken from Bitstamp.

And I would return only the price as a number and not the complete DTO. This would also make lines 14-15 unnecessary.

TonySemikin: "And I would return only the price as a number" - was expecting this comment. Tbh I always prefer to return type if you request type, instead of primitive. In this case we are interested only in amount, but generally we are getting the Price - getMatchingPrice(), so returning Price will give more type safety and flexibility imho (even though it is more lines)

"And in case of USDC the reference price should be taken from Bitstamp" - sure, lets talk through possible rules tomorrow, I was copying the existing rules and there are no difference in getPrice() case, its same base class ExchangeService, so if there are some specific rules - lets discuss or keep it YAGNI.

PR #256, opened 2022-06-14
Apply autoswap to node api

TonySemikin:
#### Pre-Release

- [ ] Check migrations
  - Login/logout
  - Registration
  - Update user data

#### Post-Release

- Test basic user operations


src/ain/node/node-client.ts
```diff
@@ -84,7 +77,7 @@ export class NodeClient {
 
   // UTXO
   get utxoFee(): number {
-    return this.chain === Chain.MAIN ? 0.00000132 : 0.0000222;
+    return this.chain === 'main' ? 0.00000132 : 0.0000222;
```
davidleomay, line 80: String in environment variable is 'mainnet'.

src/ain/node/node-client.ts
```diff
@@ -248,8 +246,8 @@ export class NodeClient {
     return { Authorization: 'Basic ' + passwordHash };
   }
 
-  private async getChain(): Promise<Chain> {
-    return this.callNode((c) => c.blockchain.getBlockchainInfo()).then((i) => i.chain as Chain);
+  private async getChain(): Promise<string> {
+    return this.callNode((c) => c.blockchain.getBlockchainInfo()).then((i) => i.chain);
   }
```
davidleomay, line 251: Can be removed completely.

src/ain/node/node.controller.ts
```diff
@@ -13,11 +13,52 @@ import { NodeService, NodeType } from './node.service';
 export class NodeController {
   constructor(private readonly nodeService: NodeService) {}
 
-  @Post(':node/:mode/rpc')
+  @Post(':node/connected/rpc')
```
davidleomay, line 16: Can we omit the 'connected' in the path (also on lines 27 and 41)? Would be nicer without ;-)

src/ain/node/node.controller.ts
```diff
@@ -13,11 +13,52 @@ import { NodeService, NodeType } from './node.service';
...
   @ApiBearerAuth()
   @ApiExcludeEndpoint()
   @UseGuards(AuthGuard(), new RoleGuard(UserRole.ADMIN))
   async rpc(@Param('node') node: NodeType, @Param('mode') mode: NodeMode, @Body() command: string): Promise<any> {
```
davidleomay, line 20: Unused parameter 'mode' (also on lines 31 and 47).

src/ain/node/node-client.ts
TonySemikin: how did it work before then? cause enum option was 'main'

src/ain/node/node-client.ts
TonySemikin: alright

src/ain/node/node-client.ts
davidleomay: Yes, that's somewhat confusing. The getInfo call returns 'main' whereas the Whale API client requires 'mainnet' 🤔 that's why in the config we have 'mainnet'.

TonySemikin: nice... :)

davidleomay (approved):

PR #347, opened 2022-08-01
[DEV-557] changed crypto route fee if a ref was used

Krysh90:
#### Pre-Release

- [x] Check migrations
  - Login/logout
  - Registration
  - Update user data

#### Post-Release

- Test basic user operations

### DIFF
--- a/src/config/config.ts
+++ b/src/config/config.ts
@@ -205,7 +205,7 @@ export class Configuration {
 
   crypto = {
     fee: 0.012,
-    feeUsedRef: 0.011,
+    refBonus: 0.001,
   };
 
   ftp = {
--- a/src/config/config.ts
+++ b/src/config/config.ts
@@ -205,6 +205,7 @@ export class Configuration {
 
   crypto = {
     fee: 0.012,
+    feeUsedRef: 0.011,
   };
 
   ftp = {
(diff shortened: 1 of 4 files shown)


src/config/config.ts
```diff
@@ -205,6 +205,7 @@ export class Configuration {
 
   crypto = {
     fee: 0.012,
+    feeUsedRef: 0.011,
```
davidleomay, line 208: I would propose to switch to a refBonus of 0.1% instead. So the fee, when using a ref code would be `crypto.fee - crypto.refBonus` (see buy fee).

Krysh90: resulting from discussion yesterday evening, changed calculation to be similar to buy routes and renamed config entry to `refBonus` and added `refBonus` to be delivered with crypto routes

src/user/models/user/user.service.spec.ts
```diff
@@ -146,4 +146,22 @@ describe('UserService', () => {
...
+    setup(AccountType.PERSONAL);
+
+    await expect(service.getUserCryptoFee(1)).resolves.toStrictEqual({ fee: 1.2, refBonus: 0 });
+  });
+
+  it('should return a fee of 1.1 and refBonus of 0.1 for crypto routes, if ref was used', async () => {
+    setup(AccountType.PERSONAL, undefined, undefined, '000-001');
+
+    await expect(service.getUserCryptoFee(1)).resolves.toStrictEqual({ fee: 1.1, refBonus: 0.1 });
+  });
```
davidleomay, line 166: Add a unit test for individual crypto fee?

Krysh90: fair enough. will add one

davidleomay (approved):

PR #350, opened 2022-08-02
[DEV-555] routes tx history

Krysh90:
#### Pre-Release

- [x] Check migrations
  - Login/logout
  - Registration
  - Update user data

#### Post-Release

- Test basic user operations


src/shared/models/fiat/__tests__/mock/fiat.entity.mock.ts
```diff
@@ -0,0 +1,14 @@
...
+  name: 'EUR',
+  enable: true,
+};
+
+export function createDefaultFiat(): Fiat {
+  return createCustomFiat({});
+}
+
+export function createCustomFiat(customValues: Partial<Fiat>): Fiat {
+  return { ...new Fiat(), ...defaultFiat, ...customValues };
```
davidleomay, line 13:
```ts
const fiat = {...new Fiat()};
fiat instanceof Fiat // false
```
Perhaps better use `Object.assign(new Fiat(), { ...defaultFiat, ...customValues })` (for all mocks)?

src/payment/models/crypto-route/crypto-route.controller.ts
```diff
@@ -54,6 +57,13 @@ export class CryptoRouteController {
     return this.cryptoRouteService.updateCrypto(jwt.id, +id, updateCryptoDto).then((b) => this.toDto(jwt.id, b));
   }
 
+  @Get(':id/history')
+  @ApiBearerAuth()
+  @UseGuards(AuthGuard(), new RoleGuard(UserRole.USER))
+  async getCryptoRouteHistory(@GetJwt() jwt: JwtPayload, @Param('id') id: string): Promise<CryptoInputHistoryDto[]> {
+    return this.cryptoInputService.getHistory(jwt.id, +id);
+  }
```
davidleomay, line 65: I think we should not return CryptoInputHistory here, but also BuyCryptoHistory. CryptoInputs are just inputs. Depending on the route, either a BuyFiat, a BuyCrypto or a CryptoStaking is created. So for CryptoRoute, you'll also find the history in the BuyCrypto table.

src/shared/models/fiat/__tests__/mock/fiat.entity.mock.ts
Krysh90: oh. wasn't aware of this. Then yes, we should definitely change it for all mocks

src/payment/models/crypto-route/crypto-route.controller.ts
Krysh90: okay, fine for me. It was more or less the only entry point i saw into crypto routes. But if another approach is more feasible and can gather more information, why not?

Krysh90:
changed crypto route history to use buy-crypto as entry point and go from there. Which took me to the point, should add a type filter to only show `BuyCrypto`, for now?

src/payment/models/buy-crypto/dto/buy-crypto-history.dto.ts
```diff
@@ -0,0 +1,12 @@
...
+export class BuyCryptoHistoryDto {
+  inputAmount: number;
+  inputAsset: string;
+  outputAmount: number;
+  outputAsset: string;
+  txId: string;
+  date: Date;
+  amlCheck: AmlCheck;
+  isComplete: boolean;
+}
```
davidleomay, line 12: All 3 history DTOs are the same. Should we use a single class/interface?

src/payment/models/buy-crypto/services/buy-crypto.service.ts
```diff
@@ -176,8 +178,52 @@ export class BuyCryptoService {
...
+      inputAmount: buyCrypto.cryptoInput.amount,
+      inputAsset: buyCrypto.cryptoInput.asset.dexName,
+      outputAmount: buyCrypto.outputAmount,
+      outputAsset: buyCrypto.outputAsset,
+      txId: buyCrypto.txId,
+      date: buyCrypto.outputDate,
+      amlCheck: buyCrypto.amlCheck,
+      isComplete: buyCrypto.isComplete,
+    };
+  }
```
davidleomay, line 225: I think you can use the `toHistoryDto` method for crypto routes as well, can't you?

davidleomay: If yes, you could also remove the cryptoInput relation in getCryptoRouteHistory.

src/payment/models/buy-fiat/buy-fiat.service.ts
```diff
@@ -86,7 +87,29 @@ export class BuyFiatService {
...
+
   // --- HELPER METHODS --- //
+  private toHistoryDto(buyFiat: BuyFiat): BuyFiatHistoryDto {
+    return {
+      inputAmount: buyFiat.inputAmount,
+      inputAsset: buyFiat.inputAsset,
+      outputAmount: buyFiat.outputAmount,
+      outputAsset: buyFiat.outputAsset,
+      txId: buyFiat.cryptoInput.inTxId,
+      date: buyFiat.cryptoInput.created,
```
davidleomay, line 107: We should use buyFiat.outputDate here, to be consistent with the other history types.

src/payment/models/buy-crypto/dto/buy-crypto-history.dto.ts
Krysh90: Thought about the very same thing, decided to have it split, as I wasn't sure, if we want to have more fields for one type of DTO. Or a completely different set of information

src/payment/models/buy-crypto/services/buy-crypto.service.ts
Krysh90: No, I can't. Judging based on the test data that is around. I need `cryptoInput` to get the real input in BTC. inside `BuyCrypto` the input amount is set in EUR and therefore weird to show, if the user has sent in BTC and suddenly we show EUR.

But might be misunderstand from my side or test data being incorrect.

src/payment/models/buy-fiat/buy-fiat.service.ts
Krysh90: I always tried to pick a date, which is in the very beginning of our processes. Not that we are showing a date to the user with some kind of processing time in between.

For example user is sending it on 1.1.22 at 14:45, we take some time to process or banks take longer and the money went out on 1.1.22 at 17:00. Then I thought we definitely should show 14:45 and not 17:00, as the action of the user was earlier

src/payment/models/buy-crypto/dto/buy-crypto-history.dto.ts
davidleomay: I think it's a good thing to have very similar history entries for different routes. It might therefore be a good idea to have multiple DTOs, but use a common base class. WDYT?

src/payment/models/buy-crypto/services/buy-crypto.service.ts
davidleomay: Oh I see. On the Dev system, BuyCryptos are not automatically filled. We therefore just entered some random information do to the test. That probably was a bad idea 🤔 

But that means that on production you'll have BTC as an input asset.

src/payment/models/buy-fiat/buy-fiat.service.ts
davidleomay: Fair enough. IMO it should be similar for all routes, and for other routes, the outputDate was taken. Perhaps it would help, if we would specifically mark it as completion date? And add the "start" date for other history types as well?

src/payment/models/buy-crypto/dto/buy-crypto-history.dto.ts
Krysh90: good point, i will extract a base class and let them implement it. if changes are needed it will be easy to extend

src/payment/models/buy-crypto/services/buy-crypto.service.ts
Krysh90: Okay that is something I am not aware. But if it is the case, I am happy to use the same function and remove the crypto one

src/payment/models/buy-fiat/buy-fiat.service.ts
Krysh90: yes consistency is key, I will change it for now and we improve in a further update.

davidleomay (approved):

PR #353, opened 2022-08-04
Feature/dev 459 ct key filter

Yannick1712:
#### Pre-Release

- [ ] Check migrations
  - Login/logout
  - Registration
  - Update user data
- [ ] **Wait for changes on payment page**

#### Post-Release

- Test basic user operations

### DIFF
--- a/src/config/config.ts
+++ b/src/config/config.ts
@@ -23,7 +23,7 @@ export class Configuration {
   defaultTwitterUrl = 'https://twitter.com/DFX_Swiss';
   defaultVolumeDecimal = 2;
   defaultPercentageDecimal = 2;
-  apiKeyVersionCT = '1'; // single digit hex number
+  apiKeyVersionCT = '0'; // single digit hex number
 
   colors = {
     white: '#FFFFFF',
--- a/src/config/config.ts
+++ b/src/config/config.ts
@@ -25,14 +25,6 @@ export class Configuration {
   defaultPercentageDecimal = 2;
   apiKeyVersionCT = '1'; // single digit hex number
 
-  HistoryFilter = {
-    buy: 0,
-    sell: 1,
-    staking: 2,
-    ref: 3,
-    lm: 4,
-  };
-
   colors = {
     white: '#FFFFFF',
     red: '#F5516C',
--- a/src/config/config.ts
+++ b/src/config/config.ts
@@ -23,7 +23,15 @@ export class Configuration {
   defaultTwitterUrl = 'https://twitter.com/DFX_Swiss';
   defaultVolumeDecimal = 2;
   defaultPercentageDecimal = 2;
-  apiKeyVersionCT = '0'; // single digit hex number
+  apiKeyVersionCT = '1'; // single digit hex number
+
+  HistoryFilter = {
+    buy: 0,
+    sell: 1,
+    staking: 2,
+    ref: 3,
+    lm: 4,
+  };
 
   colors = {
     white: '#FFFFFF',
--- a/src/shared/services/api-key.service.ts
+++ b/src/shared/services/api-key.service.ts
@@ -44,7 +44,7 @@ export class ApiKeyService {
   }
 
   public getFilter(apiKey: string): HistoryFilter {
-    const version = apiKey.substring(apiKey.length - 1, apiKey.length);
+    const version = apiKey.substring(apiKey.length - this.versionLength, apiKey.length);
 
     switch (version) {
       case '0':
--- a/src/shared/services/api-key.service.ts
+++ b/src/shared/services/api-key.service.ts
@@ -33,11 +33,11 @@ export class ApiKeyService {
 
   public getActiveFilter(apiKey: string, filter: HistoryQuery): HistoryQuery {
     const apiKeyFilter = parseInt(apiKey.substring(apiKey.length - 5, apiKey.length - 1), 16);
-    const apiKeyVersion = parseInt(apiKey.substring(apiKey.length - 1, apiKey.length));
+    const apiKeyVersion = apiKey.substring(apiKey.length - 1, apiKey.length);
 
-    if (apiKeyVersion == 0) {
+    if (apiKeyVersion == '0') {
       return filter;
-    } else if (apiKeyVersion == 1) {
+    } else if (apiKeyVersion == '1') {
       Object.entries(Config.HistoryFilter).map(
         ([key, value]) =>
           (filter[key] =
--- a/src/shared/services/api-key.service.ts
+++ b/src/shared/services/api-key.service.ts
@@ -5,6 +5,14 @@ import { HistoryQuery } from 'src/payment/models/history/dto/history-query.dto';
 import { User } from 'src/user/models/user/user.entity';
 import { Util } from '../util';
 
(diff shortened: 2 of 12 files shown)


src/config/config.ts
```diff
@@ -23,7 +23,15 @@ export class Configuration {
...
-  apiKeyVersionCT = '0'; // single digit hex number
+  apiKeyVersionCT = '1'; // single digit hex number
+
+  HistoryFilter = {
+    buy: 0,
+    sell: 1,
+    staking: 2,
+    ref: 3,
+    lm: 4,
+  };
```
davidleomay, line 34: I would place this as an enum next to api-key.service.

src/shared/services/api-key.service.ts
```diff
@@ -0,0 +1,62 @@
...
+import { Util } from '../util';
+
+@Injectable()
+export class ApiKeyService {
+  constructor() {}
+
+  public createKey(address: string, queryFilter: HistoryFilter): string {
+    let hash = Util.createHash(Util.createHash(address + new Date().toISOString(), 'sha256'), 'md5').toUpperCase();
+
+    return hash.substring(0, hash.length - 5) + this.getFilterHash(queryFilter) + Config.apiKeyVersionCT;
```
davidleomay, line 15: We should try to store this magic numbers (in this case 5) into variables.

src/shared/services/api-key.service.ts
```diff
@@ -0,0 +1,62 @@
...
+              : filter[key] != null
+              ? filter[key]
+              : null),
+      );
+
+      return filter;
+    }
+  }
+
+  private getFilterHash(queryFilter: HistoryFilter): string {
```
davidleomay, line 55: It's actually not a hash, more a code 😉

src/shared/services/api-key.service.ts
```diff
@@ -0,0 +1,62 @@
...
+    let filterHashSubString = null;
+    Object.entries(queryFilter).map(([key]) => (filterHashSubString += Math.pow(2, Number(Config.HistoryFilter[key]))));
+    filterHashSubString = '0000' + (filterHashSubString ? filterHashSubString.toString(16) : '');
+    filterHashSubString = filterHashSubString.substring(filterHashSubString.length - 4, filterHashSubString.length);
+    return filterHashSubString;
```
davidleomay, line 60:
```ts
const filterCode = Util.sum(Object.keys(queryFilter).map((key) => Math.pow(2, Config.HistoryFilter[key])));
return filterCode.toString(16).padStart(4, '0');
```

davidleomay: As we use a hex code here, 2 or 3 digits should probably be enough.

src/shared/services/api-key.service.ts
```diff
@@ -0,0 +1,62 @@
...
+          (filter[key] =
+            (apiKeyFilter & Math.pow(2, value)) > 0
+              ? ((apiKeyFilter & Math.pow(2, value)) > 0).toString()
+              : filter[key] != null
+              ? filter[key]
+              : null),
+      );
+
+      return filter;
+    }
```
davidleomay, line 52: I would remove the filter parameter and just return the HistoryQuery resulting from the parsing. In the calling function you can use object destructuring to merge the queries (`const a = {...b, ...c};`)

src/shared/services/api-key.service.ts
```diff
@@ -0,0 +1,62 @@
...
+      return filter;
+    } else if (apiKeyVersion == '1') {
+      Object.entries(Config.HistoryFilter).map(
+        ([key, value]) =>
+          (filter[key] =
+            (apiKeyFilter & Math.pow(2, value)) > 0
+              ? ((apiKeyFilter & Math.pow(2, value)) > 0).toString()
+              : filter[key] != null
+              ? filter[key]
+              : null),
```
davidleomay, line 48:
```ts
Object.entries(Config.HistoryFilter).filter(([key, value) => apiKeyFilter & Math.pow(2, value) > 0).reduce((prev, [key, value]) => {...prev, [key]: 'true', {});
```

src/user/models/user/user.service.ts
```diff
@@ -38,6 +40,7 @@ export class UserService {
     private readonly walletService: WalletService,
     private readonly settingService: SettingService,
     private readonly dfiTaxService: DfiTaxService,
+    private readonly apiKeyCT: ApiKeyService,
```
davidleomay, line 43: Rename to apiKeyService?

src/shared/services/api-key.service.ts
```diff
@@ -0,0 +1,62 @@
...
+    if (!user.apiKeyCT) throw new BadRequestException('API key is null');
+    return Util.createHash(user.apiKeyCT + user.created, 'sha256').toUpperCase();
+  }
+
+  public getSign(user: User, timestamp: string): string {
+    const secret = this.getSecret(user);
+    return Util.createHash(secret + timestamp, 'sha256').toUpperCase();
+  }
+
+  public checkKeySign(user: User, sign: string, timestamp: string): boolean {
```
davidleomay, line 28: Rename to `isValid` or similar? Would make the meaning of the result clear.

src/shared/services/api-key.service.ts
```diff
@@ -0,0 +1,62 @@
...
+    const userSign = this.getSign(user, timestamp);
+
+    return sign.toUpperCase() == userSign || Util.daysDiff(new Date(timestamp), new Date()) <= 1;
```
davidleomay, line 31: Are you sure that this is right? 😉

src/shared/services/api-key.service.ts
Yannick1712: You mean as a variable in config?

davidleomay (approved):

src/shared/services/api-key.service.ts
```diff
@@ -9,7 +9,7 @@ export class ApiKeyService {
   private versionLength = 1;
   private filterLength = 3;
   private get suffixLength(): number {
-    return this.versionLength + this.filterLength;
+    return this.versionLength;
```
davidleomay, line 12: suffixLength can be removed completely and version can be back to '0'.

src/shared/services/api-key.service.ts
```diff
@@ -38,25 +38,26 @@ export class ApiKeyService {
...
   // --- KEY HANDLING --- //
-  public create(address: string, filter: HistoryFilter): string {
+  public createKey(address: string): string {
     const hash = Util.createHash(Util.createHash(address + new Date().toISOString(), 'sha256'), 'md5').toUpperCase();
-    return hash.substring(0, hash.length - this.suffixLength) + this.filterToCode(filter) + Config.apiKeyVersionCT;
+    return hash.substring(0, hash.length - this.suffixLength) + Config.apiKeyVersionCT;
   }
 
-  public getFilter(apiKey: string): HistoryFilter {
+  public getFilter(apiKey: string, filterCode: string): HistoryFilter {
```
davidleomay, line 46: I think the filter is not depending on the (version of the) apiKey, is it?

davidleomay (approved):

PR #358, opened 2022-08-08
Feature/dev 416 btc input history

Yannick1712:
#### Pre-Release

- [ ] Check migrations
  - Login/logout
  - Registration
  - Update user data

#### Post-Release

- Test basic user operations


src/payment/models/history/history.service.ts
```diff
@@ -387,4 +487,67 @@ export class HistoryService {
...
+        r.stakingRewards.length > 1
+          ? r.stakingRewards.forEach((reward: StakingReward, index: number) =>
+              index > 0
+                ? rewards.find((origReward) => (origReward.id == reward.id ? (origReward.txId += index) : null))
+                : null,
+            )
+          : null,
+      );
+
+    return rewards;
```
davidleomay, line 508: Perhaps a more readable solution?

Place the following in util.ts

```ts
  static groupBy<T>(list: T[], key: KeyType<T, string>): Map<string, T[]> {
    return list.reduce(
      (map, item) =>
        map.set(item[key] as unknown as string, (map.get(item[key] as unknown as string) ?? []).concat(item)),
      new Map<string, T[]>(),
    );
  }
```

and use it for all duplicate checks (example below).

```ts
  private fixDuplicateTxRewards(rewards: StakingReward[]): StakingReward[] {
    const duplicates = Array.from(Util.groupBy(rewards, 'txId'))
      .map(([_, rewards]) => rewards)
      .filter((r) => r.length > 1);

    duplicates.forEach((rewards) => rewards.forEach((r, i) => (r.txId += i > 0 ? i : '')));

    return rewards;
  }
```

davidleomay (approved):

PR #364, opened 2022-08-16
[DEV-624] link changes

Krysh90:
#### Pre-Release

- [ ] Check migrations
  - Login/logout
  - Registration
  - Update user data

#### Post-Release

- Test basic user operations

### DIFF
--- a/src/user/models/link/link-address.entity.ts
+++ b/src/user/models/link/link-address.entity.ts
@@ -1,0 +1,43 @@
+import { IEntity } from 'src/shared/models/entity';
+import { Column, Entity, Generated, Index } from 'typeorm';
+
+@Entity()
+export class LinkAddress extends IEntity {
+  @Column({ length: 256 })
+  existingAddress: string;
+
+  @Column({ length: 256 })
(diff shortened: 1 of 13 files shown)


src/user/models/link/link-address.entity.ts
```diff
@@ -0,0 +1,43 @@
...
+  create(existingAddress: string, newAddress: string): this {
+    this.existingAddress = existingAddress;
+    this.newAddress = newAddress;
+
+    const tomorrow = new Date();
+    tomorrow.setDate(tomorrow.getDate() + 1);
+    this.expiration = tomorrow;
+
+    return this;
+  }
```
davidleomay, line 32: How about a static create method? WDYT?

davidleomay (approved):

src/user/models/link/link-address.entity.ts
Krysh90: I don't really mind, tried to use our new approach. A static function would be better, I think. As it is more readable. Will change it

PR #365, opened 2022-08-17
[DEV-633] added blockchain info to JWT, user and asset

Krysh90:
#### Pre-Release

- [ ] Check migrations
  - Login/logout
  - Registration
  - Update user data

#### Post-Release

- Test basic user operations

### DIFF
--- a/src/ain/services/crypto.service.ts
+++ b/src/ain/services/crypto.service.ts
@@ -3,18 +3,25 @@ import { verify } from 'bitcoinjs-message';
 import { MainNet } from '@defichain/jellyfish-network';
 import { isEthereumAddress } from 'class-validator';
 import { verifyMessage } from 'ethers/lib/utils';
-import { Blockchain } from '../node/node.service';
+
+export enum Blockchain {
+  DEFICHAIN = 'DeFiChain',
+  BITCOIN = 'Bitcoin',
+  ETHEREUM = 'Ethereum',
+}
 
 @Injectable()
 export class CryptoService {
-  public verifySignature(message: string, address: string, signature: string, blockchain: Blockchain): boolean {
+  public verifySignature(message: string, address: string, signature: string): boolean {
+    const blockchain = this.getBlockchainBasedOn(address);
+
     let isValid = false;
     try {
       isValid = this.verify(message, address, signature, blockchain);
     } catch (e) {}
 
-    if (!isValid) {
-      isValid = this.fallbackVerify(message, address, signature);
+    if (!isValid && blockchain !== Blockchain.ETHEREUM) {
+      isValid = this.fallbackVerify(message, address, signature, blockchain);
     }
     return isValid;
   }
@@ -29,7 +36,7 @@ export class CryptoService {
     return address.match(/^(bc1|[13])[a-zA-HJ-NP-Z0-9]{25,39}$/)?.length > 1 ?? false;
   }
 
-  private fallbackVerify(message: string, address: string, signature: string) {
+  private fallbackVerify(message: string, address: string, signature: string, blockchain: Blockchain) {
     let isValid = false;
     const flags = [...Array(12).keys()].map((i) => i + 31);
     for (const flag of flags) {
@@ -39,7 +46,7 @@ export class CryptoService {
       sigBuffer = Buffer.concat([flagByte, sigBuffer]);
       const candidateSig = sigBuffer.toString('base64');
       try {
-        isValid = this.verify(message, address, candidateSig, Blockchain.DEFICHAIN);
+        isValid = this.verify(message, address, candidateSig, blockchain);
         if (isValid) break;
       } catch (e) {}
     }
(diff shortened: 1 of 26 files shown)


src/ain/services/crypto.service.ts
```diff
@@ -3,13 +3,14 @@ import { verify } from 'bitcoinjs-message';
 import { MainNet } from '@defichain/jellyfish-network';
 import { isEthereumAddress } from 'class-validator';
 import { verifyMessage } from 'ethers/lib/utils';
+import { Blockchain } from '../node/node.service';
```
davidleomay, line 6: It's probably better to move the Blockchain enum to crypto.service.ts?

src/ain/services/crypto.service.ts
```diff
@@ -36,17 +39,22 @@ export class CryptoService {
       sigBuffer = Buffer.concat([flagByte, sigBuffer]);
       const candidateSig = sigBuffer.toString('base64');
       try {
-        isValid = this.verify(message, address, candidateSig);
+        isValid = this.verify(message, address, candidateSig, Blockchain.DEFICHAIN);
```
davidleomay, line 42: It's a bit weird to have the blockchain hard coded here. Either the `fallbackVerify` is only used in `verifyDefichain` (which is probably the correct solution) or the blockchain should be added as param.
I think this fallback is used to be able to verify all types of DeFiChain adresses (8xxx, dxxx, df1xxx). Seems like you found a better solution for Bitcoin. Could this also be used for DeFiChain?

src/ain/services/crypto.service.ts
```diff
@@ -3,13 +3,14 @@ import { verify } from 'bitcoinjs-message';
 import { MainNet } from '@defichain/jellyfish-network';
 import { isEthereumAddress } from 'class-validator';
 import { verifyMessage } from 'ethers/lib/utils';
+import { Blockchain } from '../node/node.service';
 
 @Injectable()
 export class CryptoService {
-  public verifySignature(message: string, address: string, signature: string): boolean {
+  public verifySignature(message: string, address: string, signature: string, blockchain: Blockchain): boolean {
```
davidleomay, line 10: Is there a reason to have blockchain as a parameter? It could be determined from the address.

src/user/models/auth/auth.service.ts
```diff
@@ -61,19 +62,19 @@ export class AuthService {
     return { accessToken: this.generateToken(user) };
   }
 
-  getSignMessage(address: string): string {
-    if (isEthereumAddress(address) || this.cryptoService.isBitcoinAddress(address))
+  getSignMessage(address: string, blockchain: Blockchain): string {
```
davidleomay, line 65: Wouldn't it be better to remove the blockchain param and determine the blockchain in this method?

src/shared/models/asset/asset.service.ts
```diff
@@ -1,13 +1,14 @@
...
 import { Asset } from './asset.entity';
 
 @Injectable()
 export class AssetService {
   constructor(private assetRepo: AssetRepository) {}
 
-  async getAllAsset(): Promise<Asset[]> {
-    return this.assetRepo.find();
+  async getAllAsset(blockchain: Blockchain): Promise<Asset[]> {
+    return this.assetRepo.find({ where: { blockchain } });
```
davidleomay, line 11: It might be a good idea to deploy this change in a later stage because all logged in users won't see any assets any more.

src/ain/services/crypto.service.ts
Krysh90: no, it was just a wrong thought I had yesterday. I have tested BTC addresses (bc1, 3, 1) and all of them worked without fallback. But the fallback should be valid for DEFICHAIN & BITCOIN, only ETHEREUM has nothing to do with that fallback mechanism. Will adjust accordingly

src/ain/services/crypto.service.ts
Krysh90: wasn't quite sure, if we want to have it like this. for me both is fine. will adapt that this function retrieves blockchain for itself

src/shared/models/asset/asset.service.ts
Krysh90: Or we are falling back to DEFICHAIN, if blockchain is not present. In that way we make sure, that nothing changes for already logged in users, and BTC / ETH addresses are getting there respective assets. wdyt?

davidleomay: Yes, makes sense 👍

src/shared/models/asset/asset.service.ts
```diff
@@ -8,6 +8,7 @@ export class AssetService {
   constructor(private assetRepo: AssetRepository) {}
 
   async getAllAsset(blockchain: Blockchain): Promise<Asset[]> {
+    if (!blockchain) blockchain = Blockchain.DEFICHAIN;
```
davidleomay, line 11: What do you think about a default parameter? Did you explicitly not use it?
```ts
  async getAllAsset(blockchain: Blockchain = Blockchain.DEFICHAIN): Promise<Asset[]> {
    return this.assetRepo.find({ where: { blockchain } });
  }
```

davidleomay (approved):

src/shared/models/asset/asset.service.ts
Krysh90: `getAllAsset` is being called with `jwt.blockchain` means, if somebody has no blockchain in there jwt yet, this function is being called with undefined, and thats' why I am checking it this way.

Tbh I am not sure, how default values behaves, if someone explicitly is calling it with `undefined`

PR #373, opened 2022-08-24
[DEV-596] Add bank tx scheduler (olkypay and frick)

Kolibri1990:
#### Pre-Release

- [ ] Check migrations
  - Login/logout
  - Registration
  - Update user data

#### Post-Release

- Test basic user operations


src/payment/models/bank-tx/frick.service.ts
```diff
@@ -0,0 +1,291 @@
...
+
+  async getAccounts(): Promise<Accounts> {
+    const url = `accounts`;
+    return await this.callApi<Accounts>(url);
+  }
+
+  // --- PARSING --- //
+  private parseTransaction(tx: Transaction): Partial<BankTx> {
+    return {
+      accountServiceRef: tx.orderId ? tx.orderId.toString() : tx.transactionNr?.toString(),
```
davidleomay, line 179: `accountServiceRef: (tx.orderId ?? tx.transactionNr)?.toString(),`

src/payment/models/bank-tx/frick.service.ts
```diff
@@ -0,0 +1,291 @@
...
+      bookingDate: tx.valutaIsExecutionDate ? new Date(tx.valuta) : new Date(tx.bookingDate),
```
davidleomay, line 180: What is this `valutaIsExecutionDate`?

src/payment/models/bank-tx/frick.service.ts
```diff
@@ -0,0 +1,291 @@
...
+    return {
+      name: account.name,
+      addressLine1: account.address,
+      city: account.city,
+      iban: account.iban,
+      memberId: account.accountNumber,
+      country: account.country,
+      bankName: account.creditInstitution,
+      creditDebitIndicator:
+        tx.direction == TransactionDirection.INCOMING ? BankTxIndicator.CREDIT : BankTxIndicator.DEBIT,
```
davidleomay, line 235: Should we get this indicator from the amount? CREDIT if positive, DEBIT if negative?

src/payment/models/bank-tx/bank-tx.service.ts
```diff
@@ -16,8 +20,38 @@ export class BankTxService {
...
+
+      if (!olkyTransactions && !frickTransactions) return;
+
+      for (const bankTx of [...olkyTransactions, ...frickTransactions]) {
+        try {
+          await this.create(bankTx);
+        } catch (e) {
+          if (!(e instanceof ConflictException)) console.error(`Failed to import transaction:`, e);
+        }
+      }
```
davidleomay, line 46: If we use the same settingKey for multiple banks, we have to break, if any request fails (because we might loose some transactions). It then could be:
```ts
const transactions = await Promise.all([
  this.olkyService.getOlkyTransactions(lastModificationTime),
  this.frickService.getFrickTransactions(lastModificationTime),
]).then(([olky, frick]) => olky.concat(frick));

for (const bankTx of transactions) {
  try {
    await this.create(bankTx);
  } catch (e) {
    if (!(e instanceof ConflictException)) console.error(`Failed to import transaction:`, e);
  }
}
```

src/payment/models/bank-tx/sepa-parser.service.ts
```diff
@@ -81,6 +81,7 @@ export class SepaParser {
           ...this.getTotalCharge(entry?.NtryDtls?.TxDtls?.Chrgs?.Rcrd),
           ...this.getRelatedPartyInfo(entry),
           ...this.getRelatedAgentInfo(entry),
+          accountIban: Config.bank.maerkiBaumann.ibanEur,
```
davidleomay, line 84: Use the IBAN from the batch (there are also multiple accounts), e.g. as method parameter.

davidleomay (changes requested):
Waiting for
- [ ] Migration
- [ ] Environment update on DEV

davidleomay (approved):

PR #385, opened 2022-08-29
[DEV-596] Add foreign information

Kolibri1990:
#### Pre-Release

- [x] Check migrations
  - Login/logout
  - Registration
  - Update user data

#### Post-Release

- Test basic user operations

### DIFF
--- a/src/payment/models/bank-tx/frick.service.ts
+++ b/src/payment/models/bank-tx/frick.service.ts
@@ -120,7 +120,7 @@ enum TransactionType {
 }
 
 enum ServiceType {
-  SWIFT = 'SWIFT ',
+  SWIFT = 'SWIFT',
   SIC = 'SIC',
   EUROSIC = 'EUROSIC',
 }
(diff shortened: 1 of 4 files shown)


src/payment/models/bank-tx/frick.service.ts
```diff
@@ -167,6 +175,10 @@ export class FrickService {
     return await this.callApi<Transactions>(`transactions`, 'GET', params);
   }
 
+  private async getTransactionAccount(orderId: string): Promise<Transactions> {
```
davidleomay, line 178: Name is a bit misleading?

davidleomay: I would have orderId parameter as number (orderId: number).

src/payment/models/bank-tx/frick.service.ts
```diff
@@ -149,6 +149,14 @@ export class FrickService {
 
     if (!transactions) return [];
 
+    for (const transaction of transactions) {
```
davidleomay, line 152: Perhaps add a comment?

davidleomay (approved):

PR #397, opened 2022-09-05
[DEV-716] changed blockchain to blockchains

Krysh90:
#### Pre-Release

- [ ] Check migrations
  - Login/logout
  - Registration
  - Update user data

#### Post-Release

- Test basic user operations


src/user/models/user/user.entity.ts
```diff
@@ -96,9 +95,6 @@ export class User extends IEntity {
   @Column({ type: 'float', default: 0 })
   stakingBalance: number;
 
-  @Column({ length: 256, default: Blockchain.DEFICHAIN })
-  blockchain: Blockchain;
-
```
davidleomay, line 101: Migration update is missing

davidleomay (changes requested):

src/user/models/user/user.entity.ts
Krysh90: good point, but we don't need to create a new migration file. we just need to adapt an existing one, as this column never got created on dev environment

davidleomay (approved):

PR #403, opened 2022-09-08
[DEV-566] use new sign message for defichain

Krysh90:
#### Pre-Release

- [ ] Check migrations
  - Login/logout
  - Registration
  - Update user data

#### Post-Release

- Test basic user operations

### DIFF
--- a/src/blockchain/ain/services/crypto.service.ts
+++ b/src/blockchain/ain/services/crypto.service.ts
@@ -10,23 +10,14 @@ import { Config } from 'src/config/config';
 export class CryptoService {
   public verifySignature(message: string, address: string, signature: string): boolean {
     const blockchains = this.getBlockchainsBasedOn(address);
-    const defichainFallbackMessage = Config.auth.signMessage + address;
 
     let isValid = false;
     try {
       isValid = this.verify(message, address, signature, blockchains);
-
-      if (!isValid && blockchains.includes(Blockchain.DEFICHAIN)) {
-        isValid = this.verify(defichainFallbackMessage, address, signature, blockchains);
-      }
     } catch (e) {}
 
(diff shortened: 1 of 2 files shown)


src/ain/services/crypto.service.ts
```diff
@@ -15,14 +16,23 @@ export enum Blockchain {
...
+        isValid = this.verify(defichainFallbackMessage, address, signature, blockchains);
+      }
     } catch (e) {}
 
     if (!isValid && !blockchains.includes(Blockchain.ETHEREUM)) {
       isValid = this.fallbackVerify(message, address, signature, blockchains);
+
+      if (!isValid && blockchains.includes(Blockchain.DEFICHAIN)) {
+        isValid = this.fallbackVerify(defichainFallbackMessage, address, signature, blockchains);
+      }
```
davidleomay, line 35: Shouldn't we put this logic into `verifyDefichain` method? Would be easier IMO.

Krysh90: yes you are 100% correct. Don't know my reasoning why I did it outside

Krysh90: ah because of the already implemented `fallbackVerify` that isn't handled within defichains verify. Hm... In which cases is this `fallbackVerify` executed? Should we in case of Defichain always execute it with the old message?

davidleomay: The `fallbackVerify` is used to support all address formats (we got this code from U-Zyn and actually never questioned or verified it 😉). So `fallbackVerify` also needs to be executed for both messages.

Krysh90: Okay just tested with a wrong and a valid signature. `fallbackVerify` does use `verify` too and that is deciding which blockchain should be used. Means that both cases have fallback to old message

davidleomay (approved):

PR #419, opened 2022-09-15
[DEV-441] UTXO minimal top-up only for user addresses

TonySemikin:
#### Pre-Release

- [x] Check migrations
  - Login/logout
  - Registration
  - Update user data

#### Post-Release

- Test basic user operations

### DIFF
--- a/src/payment/models/payout/services/payout-defichain.service.ts
+++ b/src/payment/models/payout/services/payout-defichain.service.ts
@@ -48,6 +48,10 @@ export class PayoutDeFiChainService {
     if (context === PayoutOrderContext.STAKING_REWARD) return Config.blockchain.default.intWalletAddress;
   }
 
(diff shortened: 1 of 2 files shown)


src/payment/models/payout/services/payout-defichain.service.ts
```diff
@@ -48,6 +48,10 @@ export class PayoutDeFiChainService {
     if (context === PayoutOrderContext.STAKING_REWARD) return Config.blockchain.default.intWalletAddress;
   }
 
+  isUserAddress(address: string): boolean {
+    return ['df1', 'tf1'].includes(address.slice(0, 3));
+  }
```
davidleomay, line 53: I would rename it to `isLightWalletAddress`.

davidleomay (approved):

PR #443, opened 2022-10-05
[DEV-801] & [DEV-802]: min Deposit

Yannick1712:
#### Pre-Release

- [ ] Check migrations
  - Login/logout
  - Registration
  - Update user data

#### Post-Release

- Test basic user operations

### DIFF
--- a/src/shared/util.ts
+++ b/src/shared/util.ts
@@ -1,8 +1,6 @@
 import { BinaryLike, createHash, createSign, KeyLike } from 'crypto';
 import { XMLValidator, XMLParser } from 'fast-xml-parser';
 import { readFile } from 'fs';
-import { values } from 'lodash';
-import { Config } from 'src/config/config';
 import { MinDeposit } from 'src/payment/models/deposit/dto/min-deposit.dto';
 
 type KeyType<T, U> = {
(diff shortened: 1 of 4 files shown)


davidleomay (commented):
Need to clarify if min. deposit display makes sense like this.

src/shared/util.ts
```diff
@@ -171,6 +173,16 @@ export class Util {
...
+  static getMinDeposit(outputAsset: string): MinDeposit[] {
+    const index = Object.entries(Config.blockchain.default.minTransactionVolume)
+      .filter(([key, _]) => key === outputAsset)
+      .map(([_, value]) => value);
+
+    return this.transformToMinDeposit(
+      index.length != 0 ? index[0] : Config.blockchain.default.minTransactionVolume.default,
+    );
+  }
+
```
davidleomay, line 185: I would move this to sell.controller, it's sell specific logic.

davidleomay: Maybe a bit simpler?
```ts
return this.transformToMinDeposit(index[0] ?? Config.blockchain.default.minTransactionVolume.default);
```

davidleomay: And rename `index` to `minVolume` or similar?

davidleomay (approved):

PR #446, opened 2022-10-07
[DEV-622], [DEV-833] buy crypto btc, eth tokens, bsc tokens

TonySemikin:
#### Pre-Release

- [x] Check migrations
  - Login/logout
  - Registration
  - Update user data
- [ ] Update dev environment
- [x] Update Asset table -> type on dev
#### Post-Release

- Test basic user operations


src/payment/models/dex/strategies/purchase-liquidity/purchase-liquidity.facade.ts
```diff
@@ -0,0 +1,93 @@
...
+
+  private getAlias(asset: Asset): Alias {
+    const { blockchain, category: assetCategory } = asset;
+
+    if (blockchain === Blockchain.BITCOIN) return Alias.BITCOIN;
+
+    if (blockchain === Blockchain.BINANCE_SMART_CHAIN) {
+      if (assetCategory === AssetCategory.CRYPTO) return Alias.BSC_CRYPTO;
+      if (assetCategory === AssetCategory.STOCK) return Alias.BSC_TOKEN;
+    }
```
davidleomay, line 80: Perhaps we need to discuss this. From what I saw, BSC_CRYPTO is for native BNB and BSC_TOKEN for all BEP-20 tokens, is this right? Then the differentiation based on AssetCategory is not really fitting 🤔 
Same for other facades and Ethereum blockchain.

davidleomay: It seems to be very similar for all chains. There are the native asset and multiple wrapped tokens, which need to be handled differently.

src/blockchain/shared/evm/evm-client.ts
```diff
@@ -32,6 +54,16 @@ export class EvmClient {
     return tx.hash;
   }
 
+  async sendToken(address: string, token: Asset, amount: number): Promise<string> {
+    const contract = this.getERC20Contract(token.chainId);
+    const decimals = await contract.decimals();
+    const targetAmount = ethers.utils.parseUnits(`${amount}`, decimals);
```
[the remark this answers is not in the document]
TonySemikin: agree, I'll wrap this into helper method

PR #450, opened 2022-10-12
[DEV-862] bankTxReturn update method

Yannick1712:
#### Pre-Release

- [ ] Check migrations
  - Login/logout
  - Registration
  - Update user data

#### Post-Release

- Test basic user operations

### DIFF
--- a/src/payment/models/bank-tx-return/bank-tx-return.service.ts
+++ b/src/payment/models/bank-tx-return/bank-tx-return.service.ts
@@ -23,20 +23,20 @@ export class BankTxReturnService {
   }
 
   async update(id: number, dto: UpdateBankTxReturnDto): Promise<BankTxReturn> {
-    const entity = await this.bankTxReturnRepo.findOne({ where: { id } });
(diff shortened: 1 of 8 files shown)


src/payment/models/bank-tx-return/bank-tx-return.service.ts
```diff
@@ -17,20 +13,36 @@ export class BankTxReturnService {
...
-    entity = this.bankTxReturnRepo.create({ ...bankTxReturn, chargebackBankTx });
+    const update = this.bankTxReturnRepo.create(dto);
 
-    return await this.bankTxReturnRepo.save(entity);
+    // chargeback bank tx
+    if (dto.chargebackBankTxId) {
+      update.chargebackBankTx = await this.bankTxRepo.findOne({ where: { id: dto.chargebackBankTxId } });
+      if (!update.chargebackBankTx) throw new BadRequestException('ChargebackBankTx not found');
+
+      const entity = await this.bankTxReturnRepo.findOne({
```
davidleomay, line 36: I would not name the inner variable equal to the one in the outer context, can cause confusion 🤔

davidleomay (approved):

PR #455, opened 2022-10-14
[DEV-770] Offshore kyc v2

Yannick1712:
#### Pre-Release

- [ ] Check migrations
  - Login/logout
  - Registration
  - Update user data

#### Post-Release

- Test basic user operations


src/user/models/kyc/kyc-webhook.service.ts
```diff
@@ -74,15 +75,24 @@ export class KycWebhookService {
...
-            //TODO change for KYC Update v2
-            kycStatus: KycCompleted(userData.kycStatus) ? KycWebhookStatus.FULL : KycWebhookStatus.NA,
+            kycStatus:
+              KycCompleted(userData.kycStatus) && spiderData?.chatbotResult
+                ? KycWebhookStatus.FULL
+                : KycCompleted(userData.kycStatus)
+                ? KycWebhookStatus.LIGHT
+                : userData.kycStatus === KycStatus.REJECTED
+                ? KycWebhookStatus.REJECTED
+                : KycWebhookStatus.NA,
```
davidleomay, line 85: Perhaps a separate method for this to restructure it a bit and make it more readable? E.g.
```ts
if (KycCompleted(...)) {
   return spiderData?.chatbotResult ? KycWebhookStatus.FULL : KycWebhookStatus.LIGHT;
} else if (...) {
   ...
} else {
   return KycWebhookStatus.NA;
}
```

src/user/models/kyc/kyc-webhook.service.ts
```diff
@@ -74,15 +75,24 @@ export class KycWebhookService {
...
             kycHash: userData.kycHash,
           },
           reason: reason,
         };
 
-        await this.http.post(`${walletUser.apiUrl}/kyc/update`, data, {
-          headers: { 'x-api-key': Config.lock.apiKey },
+        const apiKey = this.walletService.getApiKeyInternal(user.wallet.name);
+        if (!apiKey) throw new ConflictException(`ApiKey for wallet ${user.wallet.name} not available`);
```
davidleomay, line 92: I would throw a standard error (`throw new Error(...)`), because this method is normally called from within our API. So we don't need HTTP codes.

src/user/models/kyc/kyc.service.ts
```diff
@@ -137,9 +138,12 @@ export class KycService {
     if (!user) throw new NotFoundException('DFX user not found');
     if (!KycCompleted(user.userData.kycStatus)) throw new ConflictException('KYC required');
 
+    const apiKey = this.walletService.getApiKeyInternal(wallet.name);
+    if (!apiKey) throw new ConflictException(`ApiKey for wallet ${wallet.name} not available`);
```
davidleomay, line 142: image
=> it's probably rather an InternalServerError

src/user/models/wallet/wallet.service.ts
```diff
@@ -1,18 +1,32 @@
...
     return (await this.walletRepo.findOne(id)) ?? (await this.walletRepo.findOne(1));
   }
 
-  // TODO: remove?
-  // private verifySignature(address: string, signature: string): boolean {
-  //   const signatureMessage = Config.auth.signMessageWallet + address;
-  //   return this.cryptoService.verifySignature(signatureMessage, address, signature);
-  // }
+  async getAllExternalServices(): Promise<Wallet[]> {
+    return await this.walletRepo.find({ where: { apiUrl: Not(IsNull()), isKycClient: true } });
```
davidleomay, line 17: API URL can be null IMO. Not all external services might want to use webhooks.

src/user/models/wallet/wallet.controller.ts
```diff
@@ -0,0 +1,65 @@
...
+    const spiderData = await this.spiderRepo.findOne({ where: { userData: { id: user.userData.id } } });
+
+    return {
+      address: user.address,
+      kycStatus:
+        KycCompleted(user.userData.kycStatus) && spiderData?.chatbotResult
+          ? KycWebhookStatus.FULL
+          : KycCompleted(user.userData.kycStatus)
+          ? KycWebhookStatus.LIGHT
+          : KycWebhookStatus.NA,
```
davidleomay, line 55: This logic is shared with KycWebhookService. Maybe we can share it somehow?

src/user/models/wallet/wallet.controller.ts
```diff
@@ -0,0 +1,65 @@
...
+      kycHash: user.userData.kycHash,
+    };
+  }
+
+  private async toDto(wallet: Wallet): Promise<WalletDto> {
+    return {
+      ...wallet,
+    };
```
davidleomay, line 63: I think we need an explicit mapping here. Otherwise the API will return all fields of the wallet (I think).

src/user/models/wallet/wallet.controller.ts
```diff
@@ -0,0 +1,65 @@
...
+  private async toDtoList(wallets: Wallet[]): Promise<WalletDto[]> {
+    return Promise.all(wallets.map((b) => this.toDto(b)));
+  }
+
+  private async toKycDataDtoList(users: User[]): Promise<KycDataDto[]> {
+    return Promise.all(users.map((b) => this.toKycDataDto(b)));
+  }
+
+  private async toKycDataDto(user: User): Promise<KycDataDto> {
+    const spiderData = await this.spiderRepo.findOne({ where: { userData: { id: user.userData.id } } });
```
davidleomay, line 46: SpiderData can be included in the original SQL query I think?

src/user/models/user-data/user-data.entity.ts
```diff
@@ -178,6 +178,11 @@ export class UserData extends IEntity {
 
   @OneToOne(() => SpiderData, (c) => c.userData, { nullable: true })
   spiderData: SpiderData;
+
+  get hasExternalUser(): boolean {
+    const externalUser = this.users.find((e) => e.wallet.isKycClient === true);
+    return externalUser ? true : false;
+  }
```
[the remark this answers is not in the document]
davidleomay: Simpler
```ts
return this.users.find((e) => e.wallet.isKycClient === true) != null;
```
or
```ts
return !!this.users.find((e) => e.wallet.isKycClient === true);
```

davidleomay (approved):

PR #458, opened 2022-10-17
[DEV-864] Add part to completed requirement and also to fallback

Kolibri1990:
#### Pre-Release

- [x] Check migrations
  - Login/logout
  - Registration
  - Update user data

#### Post-Release

- Test basic user operations

### DIFF
--- a/src/user/services/spider/spider-sync.service.ts
+++ b/src/user/services/spider/spider-sync.service.ts
@@ -232,18 +232,8 @@ export class SpiderSyncService {
   // --- HELPER METHODS --- //
 
   private async getIdentPdfUrl(userData: UserData): Promise<string> {
-    let result = await this.getIdentResult(userData, KycContentType.PDF);
+    const result = await this.getIdentResult(userData, KycContentType.PDF);
 
-    if (!result) {
-      //fallback to other ident
-      result = await this.getIdentResult(
-        userData,
-        KycContentType.PDF,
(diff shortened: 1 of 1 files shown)


src/user/services/spider/spider-sync.service.ts
```diff
@@ -290,31 +290,30 @@ export class SpiderSyncService {
...
+
+      part = await this.spiderApi
+        .getDocumentVersionParts(userData.id, false, document, version.name)
+        .then((parts) => parts.find((p) => p.contentType === documentType));
     }
 
-    return { document, version: version?.name };
+    if (!version) return null;
+
+    return { document, version: version.name, part };
```
davidleomay, line 317: Use a separate method for document fetch:
```ts
  private async getCompletedIdentDocument(
    userData: UserData,
    document: KycDocument,
    contentType: KycContentType,
  ): Promise<{ document: KycDocument; version: string; part: DocumentVersionPart }> {
    const version = await this.spiderApi.getDocumentVersion(userData.id, false, document, KycDocumentState.COMPLETED);
    if (!version) return null;

    const part = await this.spiderApi
      .getDocumentVersionParts(userData.id, false, document, version.name)
      .then((parts) => parts.find((p) => p.contentType === contentType));
    if (!part) return null;

    return { document, version: version.name, part };
```
and use it like 
```ts
...
let result = await this.getCompletedIdentDocument(...);
if (!result) {
   // fallback ...
}
```

davidleomay (approved):

PR #461, opened 2022-10-18
Remove DFI-Token

davidleomay:
#### Pre-Release

- [ ] Check migrations
  - Login/logout
  - Registration
  - Update user data
- [ ] Remove `-Token` for DFI-Token asset in dev DB

#### Post-Release

- Test basic user operations

### DIFF
--- a/src/payment/models/history/history.service.ts
+++ b/src/payment/models/history/history.service.ts
@@ -464,9 +464,7 @@ export class HistoryService {
   }
 
   private getAssetSymbol(dexName: string): string {
-    // TODO: new col in asset table to differentiate stocks and crypto token?
-    dexName = dexName.replace('-Token', '');
-
+    // TODO: use col from asset table to differentiate stocks and crypto token?
     return dexName === 'DUSD'
       ? 'DUSD4'
       : ['DFI', 'BTC', 'ETH', 'BCH', 'DOGE', 'LTC', 'USDC', 'USDT'].includes(dexName)
--- a/src/shared/models/asset/asset.service.ts
+++ b/src/shared/models/asset/asset.service.ts
@@ -7,7 +7,7 @@ import { Asset, AssetType } from './asset.entity';
 export interface AssetQuery {
   dexName: string;
   blockchain: string;
-  type?: AssetType;
+  type: AssetType;
 }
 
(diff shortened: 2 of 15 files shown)


src/payment/models/history/history.service.ts
```diff
@@ -464,9 +464,7 @@ export class HistoryService {
   }
 
   private getAssetSymbol(dexName: string): string {
-    // TODO: new col in asset table to differentiate stocks and crypto token?
-    dexName = dexName.replace('-Token', '');
-
+    // TODO: use col from asset table to differentiate stocks and crypto token?
```
TonySemikin, line 467: is this comment still relevant or can it be removed?

src/shared/models/asset/asset.service.ts
```diff
@@ -2,12 +2,12 @@ import { Injectable } from '@nestjs/common';
...
 import { AssetRepository } from 'src/shared/models/asset/asset.repository';
 import { In } from 'typeorm';
-import { Asset } from './asset.entity';
+import { Asset, AssetType } from './asset.entity';
 
 export interface AssetQuery {
   dexName: string;
   blockchain: string;
-  isToken?: boolean;
+  type?: AssetType;
```
TonySemikin, line 10: Then we either need to default to Coin in getAssetByQuery (easy, temp solution) or actually make type mandatory too (harder fix, cause currently it is not always clear what type should we look for in calling contexts, like buy-crypto-batch and need some heavy refactoring)... wdyt?

src/payment/models/history/history.service.ts
davidleomay: It's still relevant. Instead of the asset list below, we could fetch the asset from database to get the category.

src/shared/models/asset/asset.service.ts
TonySemikin: defaulting to Coin actually not an option, cause it anyways will require refactoring many calling contexts to explicitly ask for Token, when its needed.

davidleomay (commented):
PR needs to wait until BuyCrypto uses asset references instead of strings.

Kolibri1990 (approved):

PR #464, opened 2022-10-20
Payment Mails to Liq Address

davidleomay:
#### Pre-Release

- [ ] Check migrations
  - Login/logout
  - Registration
  - Update user data

#### Post-Release

- Test basic user operations


davidleomay (commented):
@TonySemikin does this logic make sense or should we rather add a new option?

TonySemikin (approved):

PR #475, opened 2022-10-24
[DEV-615] Add btc fee to crypto input

Kolibri1990:
#### Pre-Release

- [x] Check migrations
  - Login/logout
  - Registration
  - Update user data

#### Post-Release

- Test basic user operations


src/payment/models/crypto-input/crypto-input.entity.ts
```diff
@@ -42,6 +42,9 @@ export class CryptoInput extends IEntity {
   @Column({ type: 'float' })
   amount: number;
 
+  @Column({ type: 'float' })
+  forwardFeeAmount: number;
```
davidleomay, line 46: This should probably be nullable: true, since it's set only after creation?

src/payment/models/crypto-input/btc-input.service.ts
```diff
@@ -165,14 +165,14 @@ export class BtcInputService extends CryptoInputService {
...
-    const outTxId = await this.btcClient.send(
+    const { outTxId, feeAmount } = await this.btcClient.send(
       address,
       input.inTxId,
       input.amount,
       input.vout,
       await this.getFeeRate(input.amount),
     );
-    await this.cryptoInputRepo.update({ id: input.id }, { outTxId });
+    await this.cryptoInputRepo.update({ id: input.id }, { outTxId, forwardFeeAmount: feeAmount });
```
davidleomay, line 175: I would do the same also for defi-input.service. It's not really important, but brings consistency.

davidleomay (approved):

PR #503, opened 2022-11-11
[DEV-845] activate users when merging

Yannick1712:
#### Pre-Release

- [ ] Check migrations
  - Login/logout
  - Registration
  - Update user data

#### Post-Release

- Test basic user operations

### DIFF
--- a/src/subdomains/generic/user/models/user-data/user-data.service.ts
+++ b/src/subdomains/generic/user/models/user-data/user-data.service.ts
@@ -225,8 +225,10 @@ export class UserDataService {
     await this.updateVolumes(slaveId);
 
     // activate users
-    for (const user of master.users) {
-      await this.userService.activateUser(user);
+    if (master.hasActiveUser) {
+      for (const user of master.users) {
+        await this.userService.activateUser(user);
+      }
     }
   }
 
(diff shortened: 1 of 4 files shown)


src/subdomains/generic/user/models/user-data/user-data.service.ts
```diff
@@ -221,6 +223,11 @@ export class UserDataService {
     // update volumes
     await this.updateVolumes(masterId);
     await this.updateVolumes(slaveId);
+
+    // activate users
+    for (const user of master.users) {
+      await this.userService.activateUser(user);
+    }
```
davidleomay, line 230: Only activate users, if one of the users was already active.

src/subdomains/generic/user/models/user-data/user-data.service.ts
```diff
@@ -221,6 +222,13 @@ export class UserDataService {
     // update volumes
     await this.updateVolumes(masterId);
     await this.updateVolumes(slaveId);
+
+    // activate users
+    if (master.hasActiveUser) {
+      for (const user of master.users) {
+        if (user?.status === UserStatus.NA) await this.userRepo.update(user.id, { status: UserStatus.ACTIVE });
```
davidleomay, line 229: Move this logic to the UserRepository and reuse it from UserService?

davidleomay (approved):

PR #534, opened 2022-11-25
[DEV-1109] LimitRequest internal mails

Yannick1712:
#### Pre-Release

- [ ] Check migrations
  - Login/logout
  - Registration
  - Update user data

#### Post-Release

- Test basic user operations

### DIFF
--- a/src/subdomains/generic/user/models/limit-request/limit-request.service.ts
+++ b/src/subdomains/generic/user/models/limit-request/limit-request.service.ts
@@ -30,7 +30,7 @@ export class LimitRequestService {
     if (!KycCompleted(user?.kycStatus)) throw new BadRequestException('KYC not yet completed');
 
     // create entity
-    const entity = this.limitRequestRepo.create(dto);
+    let entity = this.limitRequestRepo.create(dto);
     entity.userData = user;
 
     // upload document proof
@@ -54,6 +54,9 @@ export class LimitRequestService {
       );
     }
 
+    // save
+    entity = await this.limitRequestRepo.save(entity);
+
     await this.notificationService
       .sendMail({
         type: MailType.INTERNAL,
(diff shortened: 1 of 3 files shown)


src/subdomains/generic/user/models/limit-request/limit-request.service.ts
```diff
@@ -51,6 +54,18 @@ export class LimitRequestService {
...
     }
 
+    await this.notificationService
+      .sendMail({
+        type: MailType.INTERNAL,
+        input: {
+          to: 'liq@dfx.swiss',
+          subject: 'LimitRequest',
+          salutation: 'New LimitRequest',
+          body: `<p>Limit: ${entity.limit} EUR</p>Investment date: ${entity.investmentDate}<p>Fund origin: ${entity.fundOrigin}</p>`,
```
davidleomay, line 64: Do we need some user info here?

src/subdomains/generic/user/models/limit-request/limit-request.service.ts
```diff
@@ -51,6 +54,18 @@ export class LimitRequestService {
...
+        },
+      })
+      .catch((error) => console.error(`Failed to send limitRequest created mail:`, error));
+
```
davidleomay, line 68: Maybe send the notification after saving? And add the id to the error log?

src/subdomains/supporting/notification/enums/index.ts
```diff
@@ -8,6 +8,7 @@ export enum MailType {
   ERROR_MONITORING = 'ErrorMonitoring',
   USER = 'User',
   PERSONAL = 'Personal',
+  INTERNAL = 'Internal',
```
davidleomay, line 11: Is INTERNAL the right name? 😉  KYC_SUPPORT and ERROR_MONITORING are also internal.

davidleomay (approved):

PR #549, opened 2022-12-02
[DEV-1130] webhook retry

Yannick1712:
#### Pre-Release

- [ ] Check migrations
  - Login/logout
  - Registration
  - Update user data

#### Post-Release

- Test basic user operations

### DIFF
--- a/src/shared/services/http.service.ts
+++ b/src/shared/services/http.service.ts
@@ -11,7 +11,7 @@ export interface HttpError {
   };
 }
 
-export type HttpRequestConfig = AxiosRequestConfig & { tryCount?: number; delay?: number };
+export type HttpRequestConfig = AxiosRequestConfig & { tryCount?: number; retryDelay?: number };
 
 @Injectable()
 export class HttpService {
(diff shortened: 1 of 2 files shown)


src/subdomains/generic/user/models/kyc/kyc-webhook.service.ts
```diff
@@ -89,9 +92,17 @@ export class KycWebhookService {
 
         await this.http.post(`${user.wallet.apiUrl}/kyc/update`, data, {
           headers: { 'x-api-key': apiKey },
+          timeout: 5000,
+          tryCount: 3,
         });
       } catch (error) {
-        console.error(`Exception during KYC webhook (${result}) for user ${userData.id}:`, error);
```
davidleomay, line 94: I would keep the log

src/subdomains/generic/user/models/kyc/kyc-webhook.service.ts
```diff
@@ -89,9 +92,17 @@ export class KycWebhookService {
 
         await this.http.post(`${user.wallet.apiUrl}/kyc/update`, data, {
           headers: { 'x-api-key': apiKey },
+          timeout: 5000,
```
davidleomay, line 95: This is the HTTP call timeout, doesn't make much sense here 🤔

src/subdomains/generic/user/models/kyc/kyc-webhook.service.ts
```diff
@@ -89,9 +92,17 @@ export class KycWebhookService {
...
         });
       } catch (error) {
-        console.error(`Exception during KYC webhook (${result}) for user ${userData.id}:`, error);
+        await this.notificationService.sendMail({
+          type: MailType.ERROR_MONITORING,
+          input: {
+            subject: 'Webhook failed',
+            errors: [`Exception during KYC webhook (${result}) for user ${userData.id}:`, error],
+          },
+        });
```
davidleomay, line 105: Not related to this change but it would generally be good to also log (and have in mail message) the user ID (not only user data ID).

src/subdomains/generic/user/models/kyc/kyc-webhook.service.ts
davidleomay: If we want to have a retry delay, we would need to adapt `HttpService`

src/shared/services/http.service.ts
```diff
@@ -11,7 +11,7 @@ export interface HttpError {
   };
 }
 
-export type HttpRequestConfig = AxiosRequestConfig & { tryCount?: number };
+export type HttpRequestConfig = AxiosRequestConfig & { tryCount?: number; delay?: number };
```
davidleomay, line 14: `retryDelay`

davidleomay (approved):

PR #572, opened 2022-12-20
[NOTASK] re-added payout changes from lock

TonySemikin:
#### Pre-Release

- [ ] Check migrations
  - Login/logout
  - Registration
  - Update user data

#### Post-Release

- Test basic user operations

### DIFF
--- a/src/shared/utils/queue-handler.ts
+++ b/src/shared/utils/queue-handler.ts
@@ -30,7 +30,7 @@ class QueueItem<T> {
 }
 
 export class QueueHandler {
-  private readonly lock = new Lock(5);
+  private readonly lock = new Lock(1200);
   private readonly queue: QueueItem<any>[] = [];
 
   constructor(scheduler: SchedulerRegistry, private readonly timeout?: number) {
--- a/src/subdomains/supporting/payout/entities/payout-order.entity.ts
+++ b/src/subdomains/supporting/payout/entities/payout-order.entity.ts
@@ -99,6 +99,8 @@ export class PayoutOrder extends IEntity {
   }
 
   pendingPayout(payoutTxId: string) {
+    if (!payoutTxId) throw new Error('No payoutTxId provided to PayoutOrder #pendingPayout(...)');
+
     this.payoutTxId = payoutTxId;
     this.status = PayoutOrderStatus.PAYOUT_PENDING;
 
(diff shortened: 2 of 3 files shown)


src/subdomains/core/buy-crypto/process/entities/buy-crypto-fees.entity.ts
```diff
@@ -39,32 +39,53 @@ export class BuyCryptoFee extends IEntity {
...
-    this.actualPayoutFeeAmount = payoutFeeAmount;
-    this.actualPayoutFeePercent = Util.round(payoutFeeAmount / transaction.outputReferenceAmount, 8);
+  addActualPayoutFee(payoutFeeAmount: number | null, transaction: BuyCrypto): this {
+    if (payoutFeeAmount == null) {
+      this.actualPayoutFeeAmount = null;
+      this.actualPayoutFeePercent = null;
+    } else {
+      this.actualPayoutFeeAmount = payoutFeeAmount;
+      this.actualPayoutFeePercent = Util.round(payoutFeeAmount / transaction.outputReferenceAmount, 8);
+    }
```
davidleomay, line 88: Can it still be null after payout is complete?

src/subdomains/core/buy-crypto/process/entities/buy-crypto.entity.ts
```diff
@@ -229,7 +229,7 @@ export class BuyCrypto extends IEntity {
     return this;
   }
 
-  complete(payoutTxId: string, payoutFee: number): this {
+  complete(payoutTxId: string, payoutFee: number | null): this {
```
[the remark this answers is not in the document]
TonySemikin: yes, if we fail to get the price for whatever reason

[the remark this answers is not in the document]
TonySemikin: for fee conversion

src/subdomains/core/buy-crypto/process/entities/buy-crypto-fees.entity.ts
TonySemikin: same thing - yes, we can just fail to get the price for fee conversion

davidleomay: image
Ok, maybe it would make sense to not call this method, if fee is null? Do the check in BuyCrypto?

TonySemikin: no, I would rather have it explicitly set to NULL in addActualPayoutFee, than just omitting it in upper stack. Better describes intention that we want those props to be set to null, if fee is not available. imho

davidleomay (approved):

PR #586, opened 2023-01-04
[DEV-1238] add bankTx Fee col

Yannick1712:
#### Pre-Release

- [ ] Check migrations
  - Login/logout
  - Registration
  - Update user data

#### Post-Release

- Test basic user operations


src/subdomains/supporting/bank/bank-tx/bank-tx.service.ts
```diff
@@ -77,8 +77,6 @@ export class BankTxService {
     // TODO später auskommentieren
     // if (bankTx.type && bankTx.type != BankTxType.UNKNOWN) throw new ConflictException('BankTx Type already set');
 
-    bankTx.type = dto.type;
-
     switch (bankTx.type) {
```
davidleomay, line 80: I think this is not a good idea, we will create entities (BuyCrypto, BankTxReturn, or BankTxRepeat) multiple times. Bank Tx type should not be changeable IMO 🤔

davidleomay (approved):

PR #600, opened 2023-01-17
[DEV-1243] added giroCode generation and blocked CH ibans for now

Krysh90:
#### Pre-Release

- [x] Check migrations
  - Login/logout
  - Registration
  - Update user data

#### Post-Release

- Test basic user operations

### DIFF
--- a/src/subdomains/core/buy-crypto/route/buy.controller.ts
+++ b/src/subdomains/core/buy-crypto/route/buy.controller.ts
@@ -109,6 +109,7 @@ export class BuyController {
 
   private async toPaymentInfoDto(userId: number, buy: Buy, dto: GetBuyPaymentInfoDto): Promise<BuyPaymentInfoDto> {
     const bankInfo = await this.getBankInfo(buy, dto);
+    const giroCode = this.generateGiroCode(buy, bankInfo, dto);
 
     return {
       ...bankInfo,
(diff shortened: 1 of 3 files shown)


src/subdomains/core/buy-crypto/route/buy.controller.ts
```diff
@@ -109,16 +109,35 @@ export class BuyController {
...
 
+  private generateGiroCode(buy: Buy, bankInfo: BankInfoDto, dto: GetBuyPaymentInfoDto): string | undefined {
+    if (Config.giroCode.disabledCountries.includes(buy.iban.slice(0, 2))) return undefined;
+    return `
+    ${Config.giroCode.service}
+    ${Config.giroCode.version}
+    ${Config.giroCode.encoding}
+    ${Config.giroCode.transfer}
+    ${bankInfo.bic}
+    DFX AG
```
davidleomay, line 132: `${bankInfo.name}`

src/subdomains/core/buy-crypto/route/dto/buy-payment-info.dto.ts
```diff
@@ -39,4 +39,7 @@ export class BuyPaymentInfoDto extends BankInfoDto {
 
   @ApiProperty({ type: MinDeposit })
   minDeposit: MinDeposit;
+
+  @ApiProperty()
```
davidleomay, line 43: `@ApiPropertyOptional()`

src/subdomains/core/buy-crypto/route/buy.controller.ts
```diff
@@ -109,16 +109,35 @@ export class BuyController {
...
+    ${bankInfo.iban}
+    ${dto.currency?.name}${dto.amount}
```
davidleomay, line 134: Return `undefined`, if currency is not defined?

davidleomay: The currency name could theoretically be undefined, as only the id is required in the DTO. It's set in `getBankInfo`, which is a bit too hidden IMO 🤔  maybe fetch it before line 111 and give it as parameter to `getBankInfo` and `generateGiroCode`?

PR #618, opened 2023-01-20
[DEV-1284] Extended price path options

davidleomay:
#### Pre-Release

- [ ] Check migrations
  - Login/logout
  - Registration
  - Update user data

#### Post-Release

- Test basic user operations

### DIFF
--- a/src/subdomains/supporting/pricing/utils/price-step.ts
+++ b/src/subdomains/supporting/pricing/utils/price-step.ts
@@ -9,16 +9,18 @@ import { PricingUtil } from './pricing.util';
 export interface PriceStepOptions {
   from?: string | 'input';
   to?: string | 'output';
-  overwriteReferenceTo?: string;
-  fallbackPrimaryTo?: string;
-  providers?: PriceStepProviders;
+  primary?: PriceStepProviderOptions;
+  reference?: PriceStepProviderOptions;
   fixedPrice?: number;
 }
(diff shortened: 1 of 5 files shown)


src/subdomains/supporting/pricing/utils/price-step.ts
```diff
@@ -9,30 +9,28 @@ import { PricingUtil } from './pricing.util';
...
+  primary?: PriceStepProviderOptions;
+  reference?: PriceStepProviderOptions;
   fixedPrice?: number;
 }
 
-export interface PriceStepProviders {
-  primary: PriceProvider[];
-  reference: PriceProvider[];
+export interface PriceStepProviderOptions {
+  providers?: PriceProvider[];
```
TonySemikin, line 18: I think providers must be required. Once someone configures primary or reference, he needs to specify providers. For example for primary providers if nothing configured, init specification will not pass it through at server start, though compile-wise everything will look good

TonySemikin: however you default to empty lists... yeah, init will actually pass, up to you

davidleomay: Ok agreed. `primary` or `reference` are not required, but `providers` should be 👍

TonySemikin (approved):

PR #632, opened 2023-01-30
[DEV-1028] Reworked missing liquidity message

TonySemikin:
#### Pre-Release

- [x] Check migrations
  - Login/logout
  - Registration
  - Update user data

#### Post-Release

- Test basic user operations


src/subdomains/supporting/dex/strategies/check-liquidity/impl/defichain-default.strategy.ts
```diff
@@ -67,19 +67,25 @@ export class DeFiChainDefaultStrategy extends CheckLiquidityStrategy {
     const { targetAmount, availableAmount, maxPurchasableAmount, isSlippageDetected, slippageMessage, feeAmount } =
       liquidity;
 
+    const targetAvailableAmount = availableAmount > 0 ? availableAmount : 0;
+    const targetMaxPurchasableAmount = maxPurchasableAmount > 0 ? maxPurchasableAmount : 0;
```
davidleomay, line 71: Could use `const targetAvailableAmount = Math.max(availableAmount, 0);` if you like 😉  (everywhere)

davidleomay (approved):

src/subdomains/supporting/dex/strategies/check-liquidity/impl/defichain-default.strategy.ts
TonySemikin: done!

PR #638, opened 2023-02-01
[DEV-1309] Liquidity management rules auto-retry

TonySemikin:
#### Pre-Release

- [x] Check migrations
  - Login/logout
  - Registration
  - Update user data

#### Post-Release

- Test basic user operations


src/subdomains/core/liquidity-management/entities/liquidity-management-rule.entity.ts
```diff
@@ -107,6 +112,19 @@ export class LiquidityManagementRule extends IEntity {
...
+  updateRuleSettings(reactivationTime: number | undefined): this {
+    if (reactivationTime !== undefined) this.reactivationTime = reactivationTime;
+
+    return this;
+  }
+
+  shouldReactivate(): boolean {
+    return (
+      this.status === LiquidityManagementRuleStatus.PAUSED &&
+      Util.minutesDiff(this.updated, new Date()) > this.reactivationTime
```
TonySemikin, line 124: when testing locally I've got a time mismatch between DB and local API. Should not be an issue in production, since api and db would be in sync clock

davidleomay: Yes, makes sense 👍

src/subdomains/core/liquidity-management/services/liquidity-management-rule.service.ts
```diff
@@ -81,6 +84,40 @@ export class LiquidityManagementRuleService {
...
+      status: LiquidityManagementRuleStatus.PAUSED,
+      reactivationTime: Not(IsNull()),
+    });
+
+    for (const rule of rules) {
+      if (rule.shouldReactivate()) {
+        rule.reactivate();
+        await this.ruleRepo.save(rule);
+
+        console.log(`Reactivated liquidity management rule ${rule.id}`);
```
davidleomay, line 116: It would be cool to have an email notification (equivalent to 'rule paused' mail).

TonySemikin: done

davidleomay (approved):

PR #783, opened 2023-05-25
[DEV-1528]: add lightning deposit address

bernd2022:
#### Pre-Release

- [ ] Check migrations
  - Login/logout
  - Registration
  - Update user data

#### Post-Release

- Test basic user operations


thunder-tests/thunderEnvironment.json
davidleomay, line 1: Changes to `thunderEnvironment.json` should never be committed (access tokens!)

thunder-tests/thunderclient.json
davidleomay, line 1: Changes to `thunderclient.json` should only be committed, if the changes are useful. In this case I would just revert it.

PR #815, opened 2023-06-26
[DEV-1593] add txId in lightning mails

Yannick1712:
#### Pre-Release

- [ ] Check migrations
  - Login/logout
  - Registration
  - Update user data

#### Post-Release

- Test basic user operations


src/shared/i18n/de/mail.json
davidleomay, line 1: Add TX ID also to on-ramp mail

davidleomay (approved):

PR #821, opened 2023-06-28
[DEV-1546] mail refactoring

Yannick1712:
#### Pre-Release

- [ ] Check migrations
  - Login/logout
  - Registration
  - Update user data

#### Post-Release

- Test basic user operations


DFXAdmin (approved):

src/subdomains/supporting/notification/factories/mail.factory.ts
```diff
@@ -161,4 +210,57 @@ export class MailFactory {
...
+        return [
+          {
+            url: element.params?.url && {
+              link: element.params.url,
+              text: this.tNew(element.key, lang)?.split('[url:')[1]?.split(']')[0],
+            },
+            style: element.params?.style ?? MailDefaultStyle,
+            key: element.params?.url ? this.tNew(element.key, lang)?.split('[url:')[0] : this.tNew(element.key, lang),
+          },
+        ];
```
davidleomay, line 263: Maybe do the splitting with regex and in a general method?
Parser method:
```ts
private parseSpecialTag(text: string): {text: string, tag: string, value: string} | undefined {
   const match = /^(.*)\[(\w*):(\w*)\]$/.exec(text);
   return match
      ? { text: match[1], tag: match[2], value: match[3] }
      : undefined;
}
```
Usage:
```ts
default:
   const text = this.tNew(element.key, lang);
   const specialTag = this.parseSpecialTag(text);

   return {
      url: specialTag?.tag === 'url' && element.params?.url ? {
         link: element.params.url,
         text: specialTag.value,
      } : undefined,
      style: element.params?.style ?? MailDefaultStyle,
      key: specialTag?.text ?? text;
   };
```

src/subdomains/supporting/notification/templates/userNew.hbs
```diff
@@ -0,0 +1,125 @@
...
+            {{#each suffix}}
+              <tr style='font-family:{{this.style}}'>
+                <td colspan='3' align='center'>
+                  {{#if this.url}}
+                    {{this.key}}<a style='color:white' href='{{this.url.link}}'>{{this.url.text}}</a>
+                  {{/if}}
+
+                  {{#unless this.url}}
+                    {{this.key}}
+                  {{/unless}}
```
davidleomay, line 80: Can this be simplified to the following?
```hbs
{{this.key}}{{#if this.url}}<a style='color:white' href='{{this.url.link}}'>{{this.url.text}}</a>{{/if}}
```

src/subdomains/supporting/notification/factories/mail.factory.ts
davidleomay: Maybe you could also add an enum for special tags ...

davidleomay (approved):

DFXAdmin (approved):

PR #822, opened 2023-06-28
[NO-TASK] set bankTx type to GSheet when buyId is not found

Kolibri1990:
#### Pre-Release

- [ ] Check migrations
  - Login/logout
  - Registration
  - Update user data

#### Post-Release

- Test basic user operations


src/subdomains/supporting/bank/bank-tx/dto/update-bank-tx.dto.ts
```diff
@@ -30,8 +30,8 @@ export class UpdateBankTxDto {
   @IsNumber()
   accountingAmountAfterFeeChf?: number;
 
-  @IsNotEmpty()
+  @IsOptional()
```
davidleomay, line 33: Needs to be `@IsNotEmpty()` (conditional validation)

davidleomay (approved):

PR #840, opened 2023-07-19
[DEV-1630] just get buyable buy, crypto & sell routes 

Kolibri1990:
#### Pre-Release

- [ ] Check migrations
  - Login/logout
  - Registration
  - Update user data

#### Post-Release

- Test basic user operations

### DIFF
--- a/src/subdomains/core/buy-crypto/routes/buy/buy.service.ts
+++ b/src/subdomains/core/buy-crypto/routes/buy/buy.service.ts
@@ -105,8 +105,7 @@ export class BuyService {
   }
 
   async getUserBuys(userId: number): Promise<Buy[]> {
-    const buyRoutes = await this.buyRepo.findBy({ user: { id: userId } });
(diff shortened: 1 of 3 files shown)


src/subdomains/core/buy-crypto/routes/buy/buy.service.ts
```diff
@@ -105,7 +105,8 @@ export class BuyService {
   }
 
   async getUserBuys(userId: number): Promise<Buy[]> {
-    return this.buyRepo.findBy({ user: { id: userId } });
+    const buyRoutes = await this.buyRepo.findBy({ user: { id: userId } });
+    return buyRoutes.filter((b) => b.asset.buyable);
```
davidleomay, line 109: Why not do it in SQL directly?
```ts
return this.buyRepo.findBy({ user: { id: userId }, { asset: buyable: true });
```

(for all 3 cases)

davidleomay (approved):

PR #841, opened 2023-07-20
[DEV-1631] masterKey signIn

Kolibri1990:
#### Pre-Release

- [ ] Check migrations
  - Login/logout
  - Registration
  - Update user data

#### Post-Release

- Test basic user operations

### DIFF
--- a/src/subdomains/generic/user/models/auth/auth.service.ts
+++ b/src/subdomains/generic/user/models/auth/auth.service.ts
@@ -61,8 +61,12 @@ export class AuthService {
     const existingUser = await this.userRepo.getByAddress(dto.address, true);
     if (existingUser) throw new ConflictException('User already exists');
 
-    if (!(await this.verifySignature(dto.address, dto.signature, dto.key)))
+    const wallet = await this.walletRepo.findOneBy({ address: dto.address });
+
+    if (wallet.masterKey != dto.signature && !(await this.verifySignature(dto.address, dto.signature, dto.key)))
       throw new BadRequestException('Invalid signature');
+
+    if (wallet.masterKey) delete dto.signature;
 
     const ref = await this.refService.get(userIp);
     if (ref) {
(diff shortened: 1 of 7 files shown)


src/subdomains/generic/user/models/auth/auth.service.ts
```diff
@@ -61,9 +61,13 @@ export class AuthService {
     const existingUser = await this.userRepo.getByAddress(dto.address, true);
     if (existingUser) throw new ConflictException('User already exists');
 
-    if (!(await this.verifySignature(dto.address, dto.signature, dto.key)))
+    const wallet = await this.walletRepo.findOneBy({ id: dto.walletId });
+
+    if (wallet.masterKey != dto.signature && !(await this.verifySignature(dto.address, dto.signature, dto.key)))
       throw new BadRequestException('Invalid signature');
 
+    if (wallet.masterKey) delete dto.signature;
```
davidleomay, line 69: We should only delete the signature, if `wallet.masterKey === dto.signature`.

```ts
if (wallet?.masterKey === dto.signature) {
      delete dto.signature;
} else if (!(await this.verifySignature(dto.address, dto.signature, dto.key)))
      throw new BadRequestException('Invalid signature');
```

src/subdomains/generic/user/models/auth/auth.service.ts
```diff
@@ -61,9 +61,13 @@ export class AuthService {
     const existingUser = await this.userRepo.getByAddress(dto.address, true);
     if (existingUser) throw new ConflictException('User already exists');
 
-    if (!(await this.verifySignature(dto.address, dto.signature, dto.key)))
+    const wallet = await this.walletRepo.findOneBy({ id: dto.walletId });
+
+    if (wallet.masterKey != dto.signature && !(await this.verifySignature(dto.address, dto.signature, dto.key)))
```
davidleomay, line 66: Wallet can be null

davidleomay: And this will get the first wallet, if the id is undefined!

davidleomay (commented):
Add a new field `wallet: string` to CreateUserDto. Wallet can be selected either by name (new field) or id (existing walletId field).

src/subdomains/generic/user/models/auth/auth.service.ts
```diff
@@ -61,13 +61,13 @@ export class AuthService {
     const existingUser = await this.userRepo.getByAddress(dto.address, true);
     if (existingUser) throw new ConflictException('User already exists');
 
-    const wallet = await this.walletRepo.findOneBy({ id: dto.walletId });
+    const wallet = await this.walletRepo.findOneBy([{ id: dto.walletId }, { name: dto.wallet }]);
```
davidleomay: This is missing in UserService L118. Maybe add a parameter `wallet?: Wallet` to the `createUser` method? Then we don't need to search for the wallet again (except if it's null we take the default one).

davidleomay (approved):

PR #850, opened 2023-08-01
[DEV-1663] return defichain assets

Yannick1712:
#### Pre-Release

- [ ] Check migrations
  - Login/logout
  - Registration
  - Update user data

#### Post-Release

- Test basic user operations


davidleomay (changes requested):
I think this doesn't work. Adapt as discussed.

Yannick1712:
Important before merging: 

- test relation in registration services
- Add MailSendDates in prd DB for old cryptoInputReturns or disable PayInMails in config

src/subdomains/supporting/payin/entities/crypto-input.entity.ts
```diff
@@ -259,4 +278,12 @@ export class CryptoInput extends IEntity {
...
     return this;
   }
+
+  get isLightningInput(): boolean {
+    return this.asset.blockchain === Blockchain.LIGHTNING;
+  }
+
+  get amlReason(): AmlReason | 'staking' {
+    return this.route instanceof Staking ? 'staking' : AmlReason.ASSET_CURRENTLY_NOT_AVAILABLE;
+  }
```
davidleomay, line 288: I think we can add STAKING_DISCONTINUED (or similar) to AML reasons?

src/subdomains/supporting/payin/services/payin-notification.service.ts
```diff
@@ -0,0 +1,95 @@
...
+
+  @Cron(CronExpression.EVERY_5_MINUTES)
+  @Lock(1800)
+  async sendNotificationMails(): Promise<void> {
+    try {
+      if (Config.processDisabled(Process.PAY_IN_MAIL)) return;
+      await this.returnedCryptoInput();
+    } catch (e) {
+      this.logger.error('Error during buy-crypto notification:', e);
+    }
```
davidleomay, line 33: try-catch is no longer required if you have a `@Lock`

src/subdomains/core/buy-crypto/process/entities/buy-crypto.entity.ts
```diff
@@ -390,8 +390,8 @@ export class BuyCrypto extends IEntity {
   }
 
   get translationReturnMailKey(): MailTranslationKey {
-    if (!this.isCryptoCryptoTransaction) return MailTranslationKey.CRYPTO_RETURN;
-    return MailTranslationKey.FIAT_RETURN;
+    if (!this.isCryptoCryptoTransaction) return MailTranslationKey.FIAT_RETURN;
+    return MailTranslationKey.CRYPTO_RETURN;
```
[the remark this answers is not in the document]
Yannick1712: I decided a renaming or switching of the names would be more understandable 😁

src/subdomains/supporting/payin/entities/crypto-input.entity.ts
Yannick1712: Hm yeah thought about that too 🤔 
It will be not really used as amlReason, so in fact it's not an amlReason but yeah we could nevertheless add it as amlReason 🤔

src/subdomains/supporting/notification/factories/mail.factory.ts
```diff
@@ -308,7 +308,7 @@ export class MailFactory {
 
   //*** STATIC HELPER METHODS ***//
 
-  static parseMailKey(mailKey: MailTranslationKey, amlReason: AmlReason): string {
+  static parseMailKey(mailKey: MailTranslationKey, amlReason: AmlReason | 'staking'): string {
```
davidleomay, line 311: Can be reverted

davidleomay (approved):

PR #887, opened 2023-10-04
[DEV-1849] added lnurl auth

bernd2022:
#### Pre-Release

- [ ] Check migrations
  - Login/logout
  - Registration
  - Update user data

#### Post-Release

- Test basic user operations

### DIFF
--- a/src/subdomains/generic/user/models/auth/__tests__/lnurl-auth.spec.ts
+++ b/src/subdomains/generic/user/models/auth/__tests__/lnurl-auth.spec.ts
@@ -1,0 +1,337 @@
+import { ForbiddenException } from '@nestjs/common';
+import { Test, TestingModule } from '@nestjs/testing';
+import { mock } from 'jest-mock-extended';
+import { Config } from 'src/config/config';
+import { LightningHelper } from 'src/integration/lightning/lightning-helper';
+import { IpLog } from 'src/shared/models/ip-log/ip-log.entity';
+import { IpLogService } from 'src/shared/models/ip-log/ip-log.service';
+import { TestUtil } from 'src/shared/utils/test.util';
+import { Util } from 'src/shared/utils/util';
+import { AuthService } from 'src/subdomains/generic/user/models/auth/auth.service';
+import { AuthLnurlSignupDto } from 'src/subdomains/generic/user/models/auth/dto/auth-lnurl.dto';
+import { AuthResponseDto } from 'src/subdomains/generic/user/models/auth/dto/auth-response.dto';
+import { AuthCacheDto, AuthLnUrlService } from '../auth-lnurl.service';
+
+describe('LnurlAuth', () => {
+  let lnUrlAuthService: AuthLnUrlService;
+  let authServiceMock: AuthService;
+  let ipLogServiceMock: IpLogService;
+
+  let signupDto: Partial<AuthLnurlSignupDto>;
+
+  const maxLoop = 10;
+  const k1Array: string[] = [];
+
+  const checkSignature = async (signupDto) =>
+    lnUrlAuthService.checkSignature('127.0.0.1', Config.url, Object.assign(new AuthLnurlSignupDto(), { ...signupDto }));
+
+  const getStatus = (signature, k1, key) => lnUrlAuthService.getStatus(signature, k1, key);
+
+  beforeAll(async () => {
+    const config = {
+      url: 'https://test.dfx.api:12345/v0.1',
+      processDisabled: () => false,
+    };
+
+    authServiceMock = mock<AuthService>();
+    ipLogServiceMock = mock<IpLogService>();
+
+    const module: TestingModule = await Test.createTestingModule({
+      imports: [],
+      providers: [
+        TestUtil.provideConfig(config),
+        { provide: AuthService, useValue: authServiceMock },
+        { provide: IpLogService, useValue: ipLogServiceMock },
+        AuthLnUrlService,
+      ],
+      controllers: [],
+    }).compile();
+
+    lnUrlAuthService = module.get<AuthLnUrlService>(AuthLnUrlService);
+  });
+
+  beforeEach(() => {
+    const ipLog = new IpLog();
+    ipLog.result = true;
+    jest.spyOn(ipLogServiceMock, 'create').mockResolvedValue(Promise.resolve(ipLog));
+
+    signupDto = createSignupDto();
+
+    createLoginLnurl(lnUrlAuthService, maxLoop, k1Array);
+    expect(k1Array.length).toEqual(maxLoop);
+  });
+
+  afterEach(() => {
+    const internalAuthCache = (lnUrlAuthService as any).authCache as Map<string, AuthCacheDto>;
+    internalAuthCache.clear();
+
+    k1Array.splice(0, k1Array.length);
+  });
+
+  describe('Auth', () => {
+    it('checks authCache containing maxloop entries', async () => {
+      const internalAuthCache = (lnUrlAuthService as any).authCache as Map<string, AuthCacheDto>;
+      expect(internalAuthCache.size).toEqual(maxLoop);
+
+      const internalAuthCacheKeys = [...internalAuthCache.keys()];
+      expect(internalAuthCacheKeys.length).toEqual(maxLoop);
+
+      internalAuthCacheKeys.forEach((k) => {
+        const authCacheEntry = internalAuthCache.get(k);
+        expect(authCacheEntry.k1).toEqual(k);
+        expect(authCacheEntry.k1CreationTime).toBeDefined();
+        expect(authCacheEntry.accessToken).toBeUndefined();
+        expect(authCacheEntry.accessTokenCreationTime).toBeUndefined();
+      });
+    });
+
+    it('should remove entries from authCache with creation time before 5 minutes', async () => {
+      const internalAuthCache = (lnUrlAuthService as any).authCache as Map<string, AuthCacheDto>;
+      const keys = [...internalAuthCache.keys()];
+
+      const authCacheEntry0 = internalAuthCache.get(keys[0]);
+      const authCacheEntry5 = internalAuthCache.get(keys[5]);
+
+      const before6MinTime = Util.minutesBefore(6).getTime();
+
+      authCacheEntry0.k1CreationTime = before6MinTime;
+      authCacheEntry5.k1CreationTime = before6MinTime;
+
+      expect(internalAuthCache.size).toEqual(maxLoop);
+
+      lnUrlAuthService.processCleanupAuthCache();
+
+      expect(internalAuthCache.size).toEqual(maxLoop - 2);
+      expect(internalAuthCache.get(keys[0])).toBeUndefined();
+      expect(internalAuthCache.get(keys[5])).toBeUndefined();
+    });
+
+    it('should cleanup entries from authCache with access token creation time before 30 seconds', async () => {
+      const internalAuthCache = (lnUrlAuthService as any).authCache as Map<string, AuthCacheDto>;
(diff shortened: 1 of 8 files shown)


src/subdomains/generic/user/models/auth/auth-lnurl.service.ts
```diff
@@ -0,0 +1,143 @@
...
+  @Cron(CronExpression.EVERY_30_SECONDS)
+  @Lock()
+  processCleanupAccessToken() {
+    if (Config.processDisabled(Process.LNURL_AUTH_CACHE)) return;
+
+    const before30SecTime = Util.secondsBefore(30).getTime();
+
+    const entriesToBeUpdated = [...this.authCache.entries()].filter(
+      (k) => k[1].accessTokenCreationTime < before30SecTime,
+    );
```
davidleomay, line 42: Why not use `this.authCache.values()`? Removes the indexing.

src/subdomains/generic/user/models/auth/auth-lnurl.service.ts
```diff
@@ -0,0 +1,143 @@
...
+  @Cron(CronExpression.EVERY_5_MINUTES)
+  @Lock()
+  processCleanupAuthCache() {
+    if (Config.processDisabled(Process.LNURL_AUTH_CACHE)) return;
+
+    const before5MinTime = Util.minutesBefore(5).getTime();
+
+    const keysToBeDeleted = [...this.authCache.entries()]
+      .filter((k) => k[1].k1CreationTime < before5MinTime)
+      .map((k) => k[0]);
```
davidleomay, line 59: Same here?

src/subdomains/generic/user/models/auth/auth-lnurl.service.ts
```diff
@@ -0,0 +1,143 @@
...
+    const before30SecTime = Util.secondsBefore(30).getTime();
+
+    const entriesToBeUpdated = [...this.authCache.entries()].filter(
+      (k) => k[1].accessTokenCreationTime < before30SecTime,
+    );
+
+    entriesToBeUpdated.forEach((e) => {
+      e[1].accessToken = undefined;
+      e[1].accessTokenCreationTime = undefined;
+    });
```
davidleomay, line 47: These k1 are already "used" -> I would delete them.

src/subdomains/generic/user/models/auth/auth-lnurl.service.ts
```diff
@@ -0,0 +1,143 @@
...
+      .map((k) => k[0]);
+
+    keysToBeDeleted.forEach((k) => this.authCache.delete(k));
+  }
+
+  createLoginLnurl(): string {
+    const k1 = Util.createHash(randomBytes(32));
+    this.authCache.set(k1, { k1: k1, k1CreationTime: Date.now() });
+
+    return LightningHelper.encodeLnurl(`${Config.url}/lnurla?tag=login&k1=${k1}&action=login`);
```
davidleomay, line 68: Maybe it's cleaner to use the URL functionality?
```ts
const url = new URL(`${Config.url}/lnurla`);
url.searchParams.set('tag', 'login');
url.searchParams.set('k1', k1);
...
return LightningHelper.encodeLnurl(url.toString());
```

src/subdomains/generic/user/models/auth/auth-lnurl.service.ts
```diff
@@ -0,0 +1,143 @@
...
+    if ('login' !== signupDto.action) return AuthLnurlResponseDto.createError('action not found');
+
+    const k1 = signupDto.k1;
+    if (!k1) return AuthLnurlResponseDto.createError('challenge not found');
+
+    const signature = signupDto.sig;
+    if (!signature) return AuthLnurlResponseDto.createError('auth signature not found');
+
+    const key = signupDto.key;
+    if (!key) return AuthLnurlResponseDto.createError('key not found');
```
davidleomay, line 112: This can be removed (already done by class-validator).

src/subdomains/generic/user/models/auth/auth-lnurl.service.ts
```diff
@@ -0,0 +1,143 @@
...
+    return LightningHelper.encodeLnurl(`${Config.url}/lnurla?tag=login&k1=${k1}&action=login`);
+  }
+
+  async checkSignature(
+    userIp: string,
+    requestUrl: string,
+    signupDto: AuthLnurlSignupDto,
+  ): Promise<AuthLnurlResponseDto> {
+    const ipLog = await this.ipLogService.create(userIp, requestUrl, signupDto.address);
+    if (!ipLog.result) throw new ForbiddenException('The country of IP address is not allowed');
```
davidleomay, line 77: We need to verify the IP of the services, not the app. So we probably need to store the IP when creating the login LNURL 🤔

src/subdomains/generic/user/models/auth/auth-lnurl.service.ts
```diff
@@ -0,0 +1,143 @@
...
+
+    const checkSignupResponse = this.checkSignupDto(signupDto);
+    if (checkSignupResponse) return checkSignupResponse;
+
+    try {
+      const k1 = signupDto.k1;
+      const signature = signupDto.sig;
+      const key = signupDto.key;
```
davidleomay, line 85: Use destructering `const { k1, signature, key } = signupDto;`.

src/subdomains/generic/user/models/auth/auth-lnurl.service.ts
```diff
@@ -0,0 +1,143 @@
...
+
+      const authCacheEntry = this.authCache.get(k1);
+      authCacheEntry.accessToken = await this.signIn(signupDto, userIp);
+      authCacheEntry.accessTokenCreationTime = Date.now();
+
+      return AuthLnurlResponseDto.createOk();
+    } catch (e) {
+      this.logger.error('Failed to login with LNURL auth:', e);
+      return { status: AuthLnurlResponseStatus.ERROR, reason: 'invalid signature' };
+    }
```
davidleomay, line 98: Why is this error `invalid signature`?

src/subdomains/generic/user/models/auth/auth-lnurl.controller.ts
```diff
@@ -0,0 +1,32 @@
...
+
+@ApiTags('LNURL')
+@Controller('auth')
+export class AuthLnurlController {
+  constructor(private readonly lnUrlService: AuthLnUrlService) {}
+
+  @Post('lnurla')
+  async getLnurlAuth(): Promise<string> {
+    return this.lnUrlService.createLoginLnurl();
+  }
```
davidleomay, line 16: Plain string responses are annoying in a client normally handling JSON. Additionally the client also needs the k1, to be able to fetch the result. So maybe return something like:
```ts
{
   k1: string,
   lnurl: string
}
```

src/subdomains/generic/user/models/auth/auth-lnurl.controller.ts
```diff
@@ -0,0 +1,32 @@
...
+    @RealIP() ip: string,
+    @Req() req: Request,
+  ): Promise<AuthLnurlResponseDto> {
+    return this.lnUrlService.checkSignature(ip, req.url, signupDto);
+  }
+
+  @Get('lnurla/status')
+  lnurlAuthStatus(@Query('k1') k1: string, @Query('signature') signature: string, @Query('key') key: string): string {
+    return this.lnUrlService.getStatus(k1, signature, key);
+  }
```
davidleomay, line 31: Same here. Maybe return something like:
```ts
{
   isComplete: boolean,
   accessToken?: string
}
```

davidleomay: The client (in our case the services) doesn't know `signature` and `key`.

src/subdomains/generic/user/models/auth/auth-lnurl.service.ts
```diff
@@ -0,0 +1,143 @@
...
+      if (e instanceof NotFoundException) return this.authService.signUp({ ...signupDto, ...session }, userIp, true);
+      throw e;
+    });
+
+    return accessToken;
+  }
+
+  getStatus(k1: string, signature: string, key: string): string {
+    const authCacheEntry = this.authCache.get(k1);
+    if (!authCacheEntry) return '';
```
davidleomay, line 136: I would throw a `NotFoundException`, if the k1 is not found.

src/subdomains/generic/user/models/auth/auth-lnurl.service.ts
```diff
@@ -0,0 +1,143 @@
...
+
+    const verifyResult = secp256k1.verify(signature, k1, key);
+    if (!verifyResult) return '';
+
+    return authCacheEntry.accessToken ?? '';
```
davidleomay, line 141: Delete the entry after a successful login.

src/subdomains/generic/user/models/auth/__tests__/lnurl-auth.spec.ts
```diff
@@ -0,0 +1,337 @@
...
+      controllers: [],
+    }).compile();
+
+    lnUrlAuthService = module.get<AuthLnUrlService>(AuthLnUrlService);
+  });
+
+  beforeEach(() => {
+    const ipLog = new IpLog();
+    ipLog.result = true;
+    jest.spyOn(ipLogServiceMock, 'create').mockResolvedValue(Promise.resolve(ipLog));
```
davidleomay, line 56: `mockResolvedValue(ipLog)` should work as well?

src/subdomains/generic/user/models/auth/auth-lnurl.service.ts
bernd2022: Here I filter over the values and need the corresponding keys.
How does this work, if I just iterate over the values?

davidleomay: Yes, makes sense. You could use the k1 stored in the value but it's probably cleaner to use the key directly.

src/subdomains/generic/user/models/auth/auth-lnurl.service.ts
```diff
@@ -0,0 +1,136 @@
...
+  async checkSignature(
+    userIp: string,
+    requestUrl: string,
+    signupDto: AuthLnurlSignupDto,
+  ): Promise<AuthLnurlSignInResponseDto> {
+    const ipLog = await this.ipLogService.create(userIp, requestUrl, signupDto.address);
+    if (!ipLog.result) throw new ForbiddenException('The country of IP address is not allowed');
+
+    const checkSignupResponse = this.checkSignupDto(signupDto);
+    if (checkSignupResponse) return checkSignupResponse;
```
davidleomay, line 81: Delete k1 from cache

src/subdomains/generic/user/models/auth/auth-lnurl.service.ts
```diff
@@ -97,40 +113,37 @@ export class AuthLnUrlService {
...
-    signupDto.wallet = 'DFX Bitcoin';
-
+  async signIn(signupDto: AuthLnurlSignupDto, servicesIp: string): Promise<string> {
     const session = { address: signupDto.address, signature: signupDto.signature };
 
     const { accessToken } = await this.authService.signIn(session, true).catch((e) => {
-      if (e instanceof NotFoundException) return this.authService.signUp({ ...signupDto, ...session }, userIp, true);
+      if (e instanceof NotFoundException) return this.authService.signUp({ ...signupDto }, servicesIp, true);
       throw e;
     });
```
davidleomay, line 133: `this.authService.signUp(signupDto, servicesIp, true)`

davidleomay (approved):

PR #931, opened 2023-11-15
[DEV-1949] add maxTxUsage

Yannick1712:
#### Pre-Release

- [ ] Check migrations
  - Login/logout
  - Registration
  - Update user data

#### Post-Release

- Test basic user operations

### DIFF
--- a/migration/1700042209571-addMaxTxUsage.js
+++ b/migration/1700042209571-addMaxTxUsage.js
@@ -1,0 +1,23 @@
+const { MigrationInterface, QueryRunner } = require("typeorm");
+
+module.exports = class addMaxTxUsage1700042209571 {
+    name = 'addMaxTxUsage1700042209571'
+
+    async up(queryRunner) {
+        await queryRunner.query(`ALTER TABLE "dbo"."buy_crypto" ADD "usedFeeIds" nvarchar(256)`);
+        await queryRunner.query(`ALTER TABLE "dbo"."buy_fiat" ADD "usedFeeIds" nvarchar(256)`);
+        await queryRunner.query(`ALTER TABLE "dbo"."fee" ADD "usages" float NOT NULL CONSTRAINT "DF_a2132c4a6848fad89fe3980b016" DEFAULT 0`);
+        await queryRunner.query(`ALTER TABLE "dbo"."fee" ADD "maxTxUsages" float`);
+        await queryRunner.query(`ALTER TABLE "fee" ADD "txUsages" float NOT NULL CONSTRAINT "DF_d8bd14f4371b371b6861d845cd8" DEFAULT 0`);
+    }
+
+    async down(queryRunner) {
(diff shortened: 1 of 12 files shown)


src/subdomains/supporting/payment/services/fee.service.ts
```diff
@@ -134,14 +138,21 @@ export class FeeService {
...
-      return { rate: customFee.rate, fixed: customFee.fixed ?? 0, payoutRefBonus: customFee.payoutRefBonus };
+    if (customFee) {
+      await this.feeRepo.update(...customFee.increaseTxUsage());
+      return {
+        feeIds: customFee.id.toString(),
+        rate: customFee.rate,
+        fixed: customFee.fixed ?? 0,
+        payoutRefBonus: customFee.payoutRefBonus,
+      };
+    }
```
davidleomay, line 155: I think we cannot just update the usage here. This might be called by the payment info endpoint. I think we need a different method for this

src/subdomains/supporting/payment/dto/fee.dto.ts
```diff
@@ -1,4 +1,5 @@
 export class FeeDto {
+  feeIds: string; // Semicolon separated id's
```
davidleomay, line 2: There are a lot of semicolons all over the place 😄  maybe we should keep it as a number array in the DTO and only map it to string in the respective entities?

migration/1700042209571-addMaxTxUsage.js
```diff
@@ -0,0 +1,23 @@
...
+    async down(queryRunner) {
+        await queryRunner.query(`ALTER TABLE "dbo"."fee" DROP CONSTRAINT "DF_d8bd14f4371b371b6861d845cd8"`);
+        await queryRunner.query(`ALTER TABLE "dbo"."fee" DROP COLUMN "txUsages"`);
+        await queryRunner.query(`ALTER TABLE "dbo"."fee" DROP COLUMN "maxTxUsages"`);
+        await queryRunner.query(`ALTER TABLE "dbo"."fee" DROP CONSTRAINT "DF_a2132c4a6848fad89fe3980b016"`);
+        await queryRunner.query(`ALTER TABLE "dbo"."fee" DROP COLUMN "usages"`);
+        await queryRunner.query(`ALTER TABLE "dbo"."buy_fiat" DROP COLUMN "usedFeeIds"`);
+        await queryRunner.query(`ALTER TABLE "dbo"."buy_crypto" DROP COLUMN "usedFeeIds"`);
+    }
+}
```
davidleomay, line 23: Update

src/subdomains/supporting/payment/services/fee.service.ts
```diff
@@ -134,14 +140,19 @@ export class FeeService {
 
   // --- HELPER METHODS --- //
 
-  private calculateFee(fees: Fee[], userDataId?: number): FeeDto {
+  private async calculateFee(fees: Fee[], userDataId?: number): Promise<FeeDto> {
```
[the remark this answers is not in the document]
Yannick1712: No, not anymore 😅

davidleomay (approved):

PR #949, opened 2023-11-28
[DEV-1967] add cryptoCrypto preparation

Yannick1712:
#### Pre-Release

- [ ] Check migrations
  - Login/logout
  - Registration
  - Update user data

#### Post-Release

- Test basic user operations


src/subdomains/core/buy-crypto/process/services/buy-crypto-preparation.service.ts
```diff
@@ -145,7 +144,7 @@ export class BuyCryptoPreparationService {
           entity.target.asset,
           inputReferencePrice,
           entity.user.userData,
-          entity.checkoutTx ? BuyPaymentMethod.CARD : BuyPaymentMethod.BANK,
+          entity.buyPaymentMethod,
```
davidleomay, line 147: What exactly is the purpose of this change?

src/subdomains/core/buy-crypto/process/services/buy-crypto-preparation.service.ts
davidleomay, line 1: We need to take special care for `minFeeAmountFiat //inputReferenceAsset if FIAT else EUR`

PR #1010, opened 2023-12-24
[DEV-2068] extend buyCrypto volumes endpoint

Yannick1712:
#### Pre-Release

- [ ] Check migrations
  - Login/logout
  - Registration
  - Update user data

#### Post-Release

- Test basic user operations

### DIFF
--- a/src/subdomains/core/sell-crypto/process/buy-fiat.service.ts
+++ b/src/subdomains/core/sell-crypto/process/buy-fiat.service.ts
@@ -164,17 +164,26 @@ export class BuyFiatService {
   }
 
   async updateVolumes(start = 1, end = 100000): Promise<void> {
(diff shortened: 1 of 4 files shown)


src/subdomains/core/sell-crypto/process/buy-fiat.service.ts
```diff
@@ -164,17 +164,26 @@ export class BuyFiatService {
   }
 
   async updateVolumes(start = 1, end = 100000): Promise<void> {
-    const sellIds = await this.sellRepo.findBy({ id: Between(start, end) }).then((l) => l.map((b) => b.id));
-    await this.updateSellVolume(sellIds);
```
davidleomay, line 168: 100'000 as a default end value might be a bit low?

davidleomay (approved):

PR #1048, opened 2024-01-21
[DEV-2079] add amount check with user tx

Yannick1712:
#### Pre-Release

- [ ] Check migrations
  - Login/logout
  - Registration
  - Update user data

#### Post-Release

- Test basic user operations


src/subdomains/supporting/payment/services/transaction-helper.ts
```diff
@@ -260,6 +266,31 @@ export class TransactionHelper implements OnModuleInit {
...
+        ),
+      );
+
+    const buyFiats = await this.buyFiatService
+      .getUserTransactions(user.id, Util.daysBefore(1))
+      .then((buyCryptos) =>
+        buyCryptos.filter(
+          (b) => b.amlCheck === CheckStatus.PASS || b.amlCheck === CheckStatus.PENDING || b.amlCheck === null,
+        ),
+      );
```
davidleomay, line 286: Shouldn't we just use `amlCheck !== CheckStatus.FAIL`?

davidleomay, line 286: I would implement the filter und sum directly in SQL to improve performance.

src/subdomains/supporting/payment/services/transaction-helper.ts
```diff
@@ -260,6 +266,31 @@ export class TransactionHelper implements OnModuleInit {
   }
 
   // --- HELPER METHODS --- //
+  private async getDailyInputAmount(inputAmount: number, from: Asset | Fiat, user?: User): Promise<number> {
+    if (!user) return inputAmount;
```
davidleomay, line 270: Maybe rename to `getVolumeLast24h`?

davidleomay (approved):

PR #1078, opened 2024-02-08
[DEV-2183] return existing bankAccount

Yannick1712:
#### Pre-Release

- [ ] Check migrations
  - Login/logout
  - Registration
  - Update user data

#### Post-Release

- Test basic user operations

### DIFF
--- a/src/subdomains/supporting/bank/bank-account/bank-account.service.ts
+++ b/src/subdomains/supporting/bank/bank-account/bank-account.service.ts
@@ -1,4 +1,4 @@
-import { BadRequestException, ConflictException, Injectable, NotFoundException } from '@nestjs/common';
+import { BadRequestException, Injectable, NotFoundException } from '@nestjs/common';
 import { Cron, CronExpression } from '@nestjs/schedule';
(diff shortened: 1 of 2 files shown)


src/subdomains/supporting/bank/bank-account/bank-account.service.ts
```diff
@@ -56,22 +56,15 @@ export class BankAccountService {
...
+      // reactivate deleted bank account
+      existing.active = true;
     }
 
-    const bankAccount = await this.getOrCreateBankAccountInternal(dto.iban, userDataId, kycType);
+    const bankAccount = existing ? existing : await this.getOrCreateBankAccountInternal(dto.iban, userDataId, kycType);
 
-    const update = await this.updateEntity(dto, bankAccount);
-    return this.bankAccountRepo.save(update);
+    return this.updateEntity(dto, bankAccount);
```
davidleomay, line 67: Maybe just like this?
```ts
const bankAccount = existing ?? await this.getOrCreateBankAccountInternal(dto.iban, userDataId, kycType);
return this.updateEntity(dto, bankAccount);
```
and set `active = true` as default in `CreateBankAccountDto`?

davidleomay (approved):

PR #1085, opened 2024-02-09
[DEV-2205] FPS price

bernd2022:
#### Pre-Release

- [ ] Check migrations
  - Login/logout
  - Registration
  - Update user data

#### Post-Release

- Test basic user operations

### DIFF
--- a/src/subdomains/supporting/pricing/pricing.module.ts
+++ b/src/subdomains/supporting/pricing/pricing.module.ts
@@ -1,12 +1,10 @@
 import { Module } from '@nestjs/common';
 import { TypeOrmModule } from '@nestjs/typeorm';
 import { BlockchainModule } from 'src/integration/blockchain/blockchain.module';
-import { FrankencoinService } from 'src/integration/blockchain/frankencoin/frankencoin.service';
 import { SharedModule } from 'src/shared/shared.module';
 import { NotificationModule } from 'src/subdomains/supporting/notification/notification.module';
 import { ExchangeModule } from '../../../integration/exchange/exchange.module';
 import { DexModule } from '../../../subdomains/supporting/dex/dex.module';
-import { LogModule } from '../log/log.module';
 import { AssetPricingMetadata } from './domain/entities/asset-pricing-metadata.entity';
 import { PriceRule } from './domain/entities/price-rule.entity';
 import { PricingController } from './pricing.controller';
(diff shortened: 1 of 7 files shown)


src/subdomains/supporting/pricing/pricing.module.ts
```diff
@@ -48,6 +52,8 @@ import { PricingServiceNew } from './services/pricing.service.new';
     PricingDeFiChainService,
     PricingCoinGeckoService,
     PricingDexService,
+    PricingFrankencoinService,
+    FrankencoinService,
```
davidleomay, line 56: Don't provide the `FrankencoinService` in the PricingModule (will create a duplicate), it should be already imported by importing the `BlockchainModule`,

src/subdomains/supporting/pricing/pricing.module.ts
```diff
@@ -31,6 +34,7 @@ import { PricingServiceNew } from './services/pricing.service.new';
     DexModule,
     NotificationModule,
     BlockchainModule,
+    LogModule,
   ],
```
davidleomay, line 38: Why this?

src/subdomains/supporting/pricing/services/integration/pricing-frankencoin.service.ts
```diff
@@ -0,0 +1,29 @@
...
+    if (!PricingFrankencoinService.ALLOWED_ASSETS.includes(from)) throw new Error(`from asset ${from} is not allowed`);
+    if (!PricingFrankencoinService.ALLOWED_ASSETS.includes(to)) throw new Error(`to asset ${to} is not allowed`);
+
+    const fpsPrice = await this.frankencoinService.getFPSPrice();
+
+    if (from === PricingFrankencoinService.ZCHF) {
+      return Price.create(from, to, Util.round(1 / fpsPrice, 8));
+    }
+
+    return Price.create(from, to, Util.round(fpsPrice / 1, 8));
```
davidleomay, line 27: There is an invert method on the price:
```ts
const price = Price.create(from, to, Util.round(fpsPrice / 1, 8));
return from === PricingFrankencoinService.ZCHF ? price : price.invert();
```

bernd2022: price.invert does not work, because the "from" and "to" tokens where also be replaced

davidleomay (approved):

PR #1089, opened 2024-02-12
[DEV-1664] Price guarantee

Kolibri1990:
#### Pre-Release

- [ ] Check migrations
  - Login/logout
  - Registration
  - Update user data

#### Post-Release

- Test basic user operations


src/subdomains/supporting/payment/entities/transaction-request.entity.ts
```diff
@@ -0,0 +1,69 @@
...
+  maxVolume: number;
+
+  @Column({ type: 'float', nullable: false })
+  amount: number;
+
+  @Column({ nullable: true })
+  sourceId: number;
+
+  @Column({ nullable: true })
+  targetId: number;
```
davidleomay, line 35: Nullable false + type 'integer' ?

PR #1092, opened 2024-02-12
[DEV-2189] change kyc missing mail text

Yannick1712:
#### Pre-Release

- [ ] Check migrations
  - Login/logout
  - Registration
  - Update user data

#### Post-Release

- Test basic user operations

### DIFF
--- a/src/subdomains/core/buy-crypto/process/services/buy-crypto-notification.service.ts
+++ b/src/subdomains/core/buy-crypto/process/services/buy-crypto-notification.service.ts
@@ -13,7 +13,7 @@ import { NotificationService } from 'src/subdomains/supporting/notification/serv
 import { In, IsNull, Not } from 'typeorm';
 import { BuyCryptoBatch } from '../entities/buy-crypto-batch.entity';
 import { BuyCryptoAmlReasonPendingStates, BuyCryptoStatus } from '../entities/buy-crypto.entity';
-import { AmlReason } from '../enums/aml-reason.enum';
+import { AmlReason, KycAmlReasons } from '../enums/aml-reason.enum';
 import { CheckStatus } from '../enums/check-status.enum';
 import { BuyCryptoRepository } from '../repositories/buy-crypto.repository';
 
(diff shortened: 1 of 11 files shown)


src/subdomains/generic/user/models/user-data/user-data.entity.ts
```diff
@@ -360,6 +360,12 @@ export class UserData extends IEntity {
     return `${Config.frontend.services}/kyc?code=${this.kycHash}`;
   }
 
+  get dilisenseUrl(): string | undefined {
+    return this.verifiedName
+      ? `https://dilisense.com/en/search/${this.verifiedName?.split(' ')?.join('%20')}`
+      : undefined;
```
davidleomay, line 366: Use `encodeURIComponent` (see e.g. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/encodeURIComponent)

src/subdomains/core/buy-crypto/process/services/buy-crypto-notification.service.ts
```diff
@@ -246,7 +246,7 @@ export class BuyCryptoNotificationService {
 
     for (const entity of entities) {
       try {
-        if (entity.user.userData.mail) {
+        if (entity.user.userData.mail && entity.user.userData.verifiedName) {
```
davidleomay, line 249: Do we really just not send the mail, if the user has no verifiedName?

src/subdomains/core/sell-crypto/process/services/buy-fiat-notification.service.ts
```diff
@@ -201,7 +201,7 @@ export class BuyFiatNotificationService {
 
     for (const entity of entities) {
       try {
-        if (entity.sell.user.userData.mail) {
+        if (entity.sell.user.userData.mail && entity.sell.user.userData.verifiedName) {
```
davidleomay, line 204: Same here?

davidleomay (approved):

PR #1158, opened 2024-03-07
[DEV-2247] Uniswap trading

bernd2022:
#### Pre-Release

- [ ] Check migrations
  - Login/logout
  - Buy/sell payment request
  - KYC page

#### Post-Release

- Test basic user operations


src/config/config.ts
```diff
@@ -330,6 +330,15 @@ export class Configuration {
...
   };
 
+  uniswap = {
+    ethereum: {
+      zchfUsdtPool: process.env.UNISWAP_ETH_ZCHF_USDT_POOL,
+    },
+    polygon: {
+      zchfUsdtPool: process.env.UNISWAP_POLYGON_ZCHF_USDT_POOL,
+    },
+  };
```
davidleomay, line 340: The pool addresses can be calculated using the following code:
```ts
  import { Token } from '@uniswap/sdk-core';
  import { FeeAmount, Pool } from '@uniswap/v3-sdk';

  async getPoolAddress(asset1: Asset, asset2: Asset): Promise<string> {
    const token1 = await this.getToken(asset1);
    const token2 = await this.getToken(asset2);

    if (token1 instanceof Token && token2 instanceof Token) {
      return Pool.getAddress(token1, token2, FeeAmount.LOWEST);
    } else {
      throw new Error(`Only tokens can be in a pool`);
    }
  }
```

src/subdomains/core/trading/controllers/trading-order.controller.ts
davidleomay, line 1: I think we don't need any controllers

src/subdomains/core/trading/controllers/trading-rule.controller.ts
davidleomay, line 1: Same here

src/subdomains/core/trading/dto/output/mappers/trading-order-output-dto.mapper.ts
davidleomay, line 1: Same for mappers and DTOs

src/subdomains/core/trading/services/trading-uniswap.service.ts
```diff
@@ -0,0 +1,265 @@
...
+      source2: {
+        name: tradingRule.source2,
+        leftAsset: tradingRule.leftAsset2,
+        rightAsset: tradingRule.rightAsset2,
+        price: price2.price,
+      },
+      swap: {
+        priceImpact: 0,
+      },
+    };
```
davidleomay, line 99: Do we need all this information? 🤔

src/subdomains/core/trading/services/trading-uniswap.service.ts
```diff
@@ -0,0 +1,265 @@
...
+
+    if (price1.isValid && price2.isValid) {
+      const ratio = price1.price / price2.price;
+      const priceImpact = (ratio - 1) * 100;
```
davidleomay, line 103: Always use factors, not percentages (0.012 instead of 1.2%)

src/subdomains/core/trading/entities/trading-order.entity.ts
```diff
@@ -0,0 +1,96 @@
...
+  @ManyToOne(() => Asset, { nullable: false, eager: true })
+  @JoinColumn()
+  assetIn: Asset;
+
+  @ManyToOne(() => Asset, { nullable: false, eager: true })
+  @JoinColumn()
+  assetOut: Asset;
+
+  @Column({ type: 'float', nullable: false })
+  amountIn: number;
```
davidleomay, line 35: Add amountOut?

PR #1194, opened 2024-03-19
[DEV-2378] amlCheck mail block

Yannick1712:
#### Pre-Release

- [ ] Check migrations
  - Login/logout
  - Buy/sell payment request
  - KYC page

#### Post-Release

- Test basic user operations


Yannick1712:
Add migration

src/subdomains/core/aml/aml.service.ts
```diff
@@ -46,6 +49,10 @@ export class AmlService {
     if (!entity.userData.lastNameCheckDate) errors.push(AmlError.NO_NAME_CHECK);
     if (Util.daysDiff(entity.userData.lastNameCheckDate) > Config.amlCheckLastNameCheckValidity)
       errors.push(AmlError.OUTDATED_NAME_CHECK);
+    if (
+      blacklist.some((b) => b.type === SpecialExternalAccountType.BANNED_MAIL && entity.userData.mail.includes(b.value))
+    )
+      errors.push(AmlError.SUSPICIOUS_MAIL);
```
davidleomay, line 55: Maybe we should use Regex here? Then we can remove the `hasSuspiciousMail` check in the user-data.

davidleomay: Mail might be undefined?

src/subdomains/core/aml/aml.service.ts
```diff
@@ -103,43 +112,64 @@ export class AmlService {
 
       if (entity.bankTx) {
         // bank
-        if (blacklist.some((b) => b.bic && b.bic === entity.bankTx.bic)) errors.push(AmlError.BIC_BLACKLISTED);
-        if (blacklist.some((b) => b.iban && b.iban === entity.bankTx.iban)) errors.push(AmlError.IBAN_BLACKLISTED);
+        if (
+          blacklist.some(
+            (b) => b.type === SpecialExternalAccountType.BANNED_BIC && new RegExp(b.value).test(entity.bankTx.bic),
+          )
```
davidleomay, line 118: Maybe we should create a helper method in the SpecialExternalAccount entity?
```ts
matches(type: SpecialExternalAccountType, value: string): boolean {
  return this.type === type && new RegExp(this.value).test(value);
}
```

davidleomay (approved):

PR #1215, opened 2024-03-27
[DEV-2258] Added new transaction endpoints

davidleomay:
#### Pre-Release

- [ ] Check migrations
  - Login/logout
  - Buy/sell payment request
  - KYC page

#### Post-Release

- Test basic user operations

### DIFF
--- a/src/subdomains/core/history/controllers/transaction.controller.ts
+++ b/src/subdomains/core/history/controllers/transaction.controller.ts
@@ -1,14 +1,4 @@
-import {
-  Controller,
-  Get,
-  NotFoundException,
-  Post,
-  Query,
-  Res,
-  Response,
-  StreamableFile,
-  UseGuards,
-} from '@nestjs/common';
+import { Controller, Get, NotFoundException, Post, Query, Res, StreamableFile, UseGuards } from '@nestjs/common';
 import { AuthGuard } from '@nestjs/passport';
 import {
   ApiBearerAuth,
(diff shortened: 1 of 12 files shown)


src/subdomains/generic/user/models/bank-data/bank-data.service.ts
```diff
@@ -54,6 +54,10 @@ export class BankDataService {
     return this.bankDataRepo.findOne({ where: { id }, relations: { userData: true } });
   }
 
+  async getBankDatasForUser(userDataId: number): Promise<BankData[]> {
+    return this.bankDataRepo.findBy({ userData: { id: userDataId }, active: true });
```
Yannick1712, line 58: Is that working? Missing userData relation

src/subdomains/core/history/controllers/transaction.controller.ts
```diff
@@ -18,41 +8,117 @@ import {
...
+
+  @Get('unassigned')
+  @ApiBearerAuth()
+  @UseGuards(AuthGuard(), new RoleGuard(UserRole.USER))
+  @ApiOkResponse({ type: UnassignedTransactionDto, isArray: true })
+  async getUnassignedTransactions(@GetJwt() jwt: JwtPayload): Promise<UnassignedTransactionDto[]> {
+    const user = await this.userService.getUser(jwt.id, { userData: true });
+    const bankDatas = await this.bankDataService.getBankDatasForUser(user.userData.id);
+
+    const txList = await this.bankTxService.getUnassignedBankTx(bankDatas.map((bd) => bd.iban));
```
Yannick1712, line 109: We will not find all bankTx with it. Theoretically we have to revert the senderAccount Method in bankTx entity and call it on the bankData iban 🤔

davidleomay: Good point. For now it should only be used for transactions with real IBANs. We will need to find a solution for all transactions later.

Yannick1712 (approved):

PR #1218, opened 2024-03-28
Improved transactions

davidleomay:
#### Pre-Release

- [ ] Check migrations
  - Login/logout
  - Buy/sell payment request
  - KYC page

#### Post-Release

- Test basic user operations

### DIFF
--- a/src/subdomains/core/history/controllers/transaction.controller.ts
+++ b/src/subdomains/core/history/controllers/transaction.controller.ts
@@ -1,4 +1,16 @@
-import { Controller, Get, NotFoundException, Post, Query, Res, StreamableFile, UseGuards } from '@nestjs/common';
+import {
+  ConflictException,
+  Controller,
+  ForbiddenException,
+  Get,
+  NotFoundException,
+  Param,
+  Post as Put,
+  Query,
+  Res,
+  StreamableFile,
+  UseGuards,
+} from '@nestjs/common';
 import { AuthGuard } from '@nestjs/passport';
 import {
   ApiBearerAuth,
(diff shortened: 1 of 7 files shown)


src/subdomains/core/history/controllers/transaction.controller.ts
```diff
@@ -62,46 +80,57 @@ export class TransactionController {
...
+  async getCsvCT(
+    @Query() query: HistoryQueryUser,
+    @Res({ passthrough: true }) res: Response,
+  ): Promise<CoinTrackingCsvHistoryDto[] | StreamableFile> {
+    if (!query.format) query.format = ExportFormat.CSV;
+    return this.getHistoryData(query, ExportType.COIN_TRACKING, res);
+  }
+
+  @Get('ChainReport')
+  @ApiOkResponse({ status: 200, type: ChainReportCsvHistoryDto, isArray: true })
```
Yannick1712, line 95: Why `status: 200` for CR but not for CT?

src/subdomains/core/history/controllers/transaction.controller.ts
```diff
@@ -113,57 +142,68 @@ export class TransactionController {
...
+    @Query('buyId') buyId: string,
+  ): Promise<void> {
+    const transaction = await this.transactionService.getTransaction(+id, { bankTx: true });
+    if (!transaction.bankTx) throw new NotFoundException('Transaction not found');
+    if (!BankTxTypeUnassigned(transaction.bankTx.type)) throw new ConflictException('Transaction already assigned');
+
+    const user = await this.userService.getUser(jwt.id, { userData: true });
+    const bankDatas = await this.bankDataService.getBankDatasForUser(user.userData.id);
+    if (!bankDatas.some((b) => b.iban === transaction.bankTx.iban))
+      throw new ForbiddenException('You can only assign your own transaction');
```
Yannick1712, line 176: Message could be a little bit confusing if the transaction is made from a new iban and the bankData is not created or with bankData iban/name special case 🤔

src/subdomains/core/history/controllers/transaction.controller.ts
```diff
@@ -113,57 +142,68 @@ export class TransactionController {
...
+
+    await this.bankTxService.update(transaction.bankTx.id, { type: BankTxType.BUY_CRYPTO, buyId: +buyId });
```
Yannick1712, line 178: buyId should be checked before, or? So if the buyId is existing and if the userData is the same with jwt userData

src/subdomains/core/history/controllers/transaction.controller.ts
davidleomay: The unassigned transaction will only appear, if the IBAN matches, so this case should never happen. And the endpoint is not publicly visible, only for DFX Services.

Yannick1712 (approved):

PR #1285, opened 2024-05-07
[DEV-2533] no cached blockchainFee for payment processing

Yannick1712:
#### Pre-Release

- [ ] Check migrations
  - Login/logout
  - Buy/sell payment request
  - KYC page

#### Post-Release

- Test basic user operations

### DIFF
--- a/src/subdomains/supporting/payment/services/fee.service.ts
+++ b/src/subdomains/supporting/payment/services/fee.service.ts
@@ -230,7 +230,8 @@ export class FeeService {
       const feeAmount = allowCached
         ? await this.blockchainFeeRepo.findOneCachedBy(`${active.id}`, where).then((fee) => fee?.amount)
         : await this.calculateBlockchainFee(active, false);
-      if (!feeAmount && !allowCached) throw new Error(`No blockchain fee found for asset ${active.id}`);
+      if (!feeAmount && !allowCached)
+        throw new Error(`No blockchain fee found for asset ${active.id} and allowCached: ${allowCached}`);
 
       return feeAmount ?? this.getBlockchainMaxFee(active.blockchain);
     } else {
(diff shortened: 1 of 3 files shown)


src/subdomains/supporting/payment/services/transaction-helper.ts
```diff
@@ -173,7 +173,7 @@ export class TransactionHelper implements OnModuleInit {
     to: Active,
     paymentMethodIn: PaymentMethod,
     paymentMethodOut: PaymentMethod,
-    allowExpiredPrice: boolean,
+    allowCachedBlockchainFee: boolean,
```
davidleomay, line 176: Here we still want to select, if the price can be expired or not. For payment info calls, we always use cached blockchain fees.

src/subdomains/supporting/payment/services/fee.service.ts
```diff
@@ -335,6 +339,19 @@ export class FeeService {
...
+    chf ??= await this.fiatService.getFiatByName('CHF');
+
+    try {
+      const { amount } = await this.payoutService.estimateBlockchainFee(asset);
+      const price = await this.pricingService.getPrice(asset, chf, allowExpiredPrice);
+
+      return price.convert(amount);
+    } catch (e) {
+      this.logger.error(`Failed to get blockchain fee of asset id ${asset.id}:`, e);
+    }
```
davidleomay, line 352: I think we can/should remove this try-catch?

src/subdomains/supporting/payment/services/fee.service.ts
```diff
@@ -213,22 +213,27 @@ export class FeeService {
...
         updated: MoreThan(Util.minutesBefore(FeeValidityMinutes)),
-      });
-      if (!fee && !allowFallback) throw new Error(`No blockchain fee found for asset ${active.id}`);
+      };
+
+      const feeAmount = allowCached
+        ? await this.blockchainFeeRepo.findOneCachedBy(`${active.id}`, where).then((fee) => fee?.amount)
+        : await this.calculateBlockchainFee(active, false);
+      if (!feeAmount && !allowCached)
+        throw new Error(`No blockchain fee found for asset ${active.id} and allowCached: ${allowCached}`);
```
davidleomay, line 234: It's not really useful to add `allowCached` to the message, it's always false in this case 😉

davidleomay (approved):

PR #1399, opened 2024-06-20
[DEV-2702] card tx limit check

Yannick1712:
#### Pre-Release

- [ ] Check migrations
  - Login/logout
  - Buy/sell payment request
  - KYC page

#### Post-Release

- Test basic user operations


davidleomay (approved):

src/subdomains/supporting/payment/services/transaction-helper.ts
```diff
@@ -437,14 +423,34 @@ export class TransactionHelper implements OnModuleInit {
...
+    let kycLimit = 0;
+    if (user?.userData.kycLevel < KycLevel.LEVEL_50) {
+      const volume24h = await this.getVolumeSince(Util.daysBefore(1), new Date(), [user]);
+      kycLimit = user?.userData.availableTradingLimit - volume24h;
+    } else {
+      kycLimit = user?.userData.availableTradingLimit ?? Number.MAX_VALUE;
+    }
+
+    return { kycLimit, defaultLimit };
+  }
```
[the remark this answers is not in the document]
davidleomay: Good question 😆

src/subdomains/supporting/payment/services/transaction-helper.ts
```diff
@@ -428,18 +428,17 @@ export class TransactionHelper implements OnModuleInit {
     paymentMethodOut: PaymentMethod,
     user: User,
   ): Promise<{ kycLimit: number; defaultLimit: number }> {
+    const volume24h =
+      user?.userData.kycLevel < KycLevel.LEVEL_50
```
Yannick1712, line 432: Is `user` optional? 🤔  In the parameter above it is not optional, but maybe the userData relation could be missing. I would add a `?` behind userData or?

src/subdomains/supporting/payment/services/transaction-helper.ts
```diff
@@ -428,18 +428,17 @@ export class TransactionHelper implements OnModuleInit {
...
+        ? await this.getVolumeSince(Util.daysBefore(1), new Date(), [user])
```
Yannick1712, line 433: Maybe we should adapt like in amlCheck `await this.getVolumeSince(Util.daysBefore(1), Util.daysAfter(1), [user])` or do you think we don't need that because edge case and only for PaymentInfos? 🤔 
I think one day in future should be not need much more resources than currentDate or?

davidleomay: The `Util.daysAfter` in AML check is based on `transaction.created` because a transaction can be quite "old". Here it's always the current day and I hope we'll never have future transactions 😆

src/subdomains/supporting/payment/services/transaction-helper.ts
davidleomay: True, `user` should be optional.

src/subdomains/supporting/payment/services/transaction-helper.ts
Yannick1712: Hehe jap your right 😁

PR #1425, opened 2024-07-01
[DEV-2506] add limitRequest in supportIssue

Yannick1712:
#### Pre-Release

- [ ] Check migrations
  - Login/logout
  - Buy/sell payment request
  - KYC page

#### Post-Release

- Test basic user operations


src/subdomains/supporting/support-issue/services/support-issue.service.ts
```diff
@@ -29,8 +31,23 @@ export class SupportIssueService {
...
     private readonly messageRepo: SupportMessageRepository,
     private readonly supportIssueNotificationService: SupportIssueNotificationService,
+    private readonly limitRequestService: LimitRequestService,
   ) {}
 
+  async createIssueInternal(userData: UserData, dto: CreateSupportIssueInternalDto): Promise<void> {
+    const newIssue = this.supportIssueRepo.create({ userData, ...dto });
+
+    const existingIssue = await this.supportIssueRepo.findOneBy({
+      userData: { id: userData.id },
```
davidleomay, line 41: `.existsBy`

src/subdomains/supporting/support-issue/services/limit-request-job.service.ts
```diff
@@ -0,0 +1,40 @@
...
+
+    const entities = await this.limitRequestRepo.find({
+      where: { supportIssue: IsNull() },
+      relations: { userData: true },
+    });
+
+    for (const entity of entities) {
+      await this.supportIssueService.createIssueInternal(entity.userData, {
+        name: entity.userData.firstname ?? '-',
+        type: SupportIssueType.LIMIT_REQUEST,
```
davidleomay, line 33: `completeName`?

src/subdomains/supporting/support-issue/services/support-issue.service.ts
```diff
@@ -52,6 +69,9 @@ export class SupportIssueService {
       newIssue.additionalInformation = dto.transaction;
     }
 
+    // limit request
+    if (dto.limitRequest) await this.limitRequestService.increaseLimitInternal(dto.limitRequest, userData);
```
davidleomay, line 73: Create the limit request with support issue? Or is this updated when removing the limit request API?

src/subdomains/supporting/support-issue/services/limit-request-job.service.ts
Yannick1712: We always use the "du"/firstname so thats the reason I used the firstname and not the completeName. But we can change it here if you want to 👍

src/subdomains/supporting/support-issue/services/support-issue.service.ts
Yannick1712: Yes right, we create a new limitRequest when creating a limitRequestSupportIssue. I think I don't understand your question 😅 
Why should it be updated when removing the limitRequest Endpoints? The limitRequest entity/service will still be used when the endpoints are removed

src/subdomains/supporting/support-issue/services/limit-request-job.service.ts
davidleomay: Ok, I just thought that most users will probably enter their full name when creating a support issue.

src/subdomains/supporting/support-issue/services/limit-request.service.ts
```diff
@@ -25,31 +25,36 @@ export class LimitRequestService {
...
         `${Util.isoDateTime(new Date())}_limit-request_user-upload_${dto.documentProofName}`,
         buffer,
         contentType as ContentType,
       );
+
+      return void this.increaseLimitInternal({ ...dto, documentProofUrl }, user);
     }
 
+    return void this.increaseLimitInternal(dto, user);
+  }
```
davidleomay, line 50: I think this doesn't wait 🤔  just use `await this.increaseLimitInternal(dto, user);`

davidleomay (approved):

PR #1460, opened 2024-07-12
[DEV-2794] sift updates

Kolibri1990:
#### Pre-Release

- [ ] Check migrations
  - Login/logout
  - Buy/sell payment request
  - KYC page

#### Post-Release

- Test basic user operations


src/subdomains/supporting/fiat-payin/services/fiat-payin-sync.service.ts
```diff
@@ -60,6 +60,7 @@ export class FiatPayInSyncService {
         const checkoutTx = await this.createCheckoutTx(refundedPayment);
         if (checkoutTx?.status === CheckoutPaymentStatus.REFUNDED) {
           await this.siftService.createChargeback({
+            $user_id: checkoutTx.buyCrypto.user.id.toString(),
             $transaction_id: checkoutTx.transaction.id.toString(),
```
davidleomay, line 64: Use `checkoutTx.transaction.user` and make sure to load it in the DB call

davidleomay: And add some `?` in case there is no user assigned to the TX

davidleomay (approved):

PR #1478, opened 2024-07-17
[DEV-2791] delete account

lapatric:
#### Pre-Release

- [ ] Check migrations
  - Login/logout
  - Buy/sell payment request
  - KYC page

#### Post-Release

- Test basic user operations


src/subdomains/generic/user/models/user/dto/update-user.dto.ts
```diff
@@ -24,4 +55,18 @@ export class UpdateUserDto {
...
+  @ValidateNested()
+  @Type(() => PartialWalletDto)
+  walletName?: PartialWalletDto;
+
+  @ApiPropertyOptional({ type: [PartialUserDto] })
+  @IsOptional()
+  @IsObject()
+  @ValidateNested()
+  @Type(() => PartialUserDto)
+  users?: PartialUserDto[];
```
davidleomay, line 71: Update as discussed

Yannick1712 (commented):
Add migration

src/shared/models/fiat/fiat.service.ts
```diff
@@ -33,4 +34,18 @@ export class FiatService {
...
+  }
+
+  ipCountryToCurrency: { [key: string]: string } = {
+    DE: 'EUR',
+    AT: 'EUR',
+    CH: 'CHF',
+    LI: 'CHF',
+    IT: 'EUR',
+    FR: 'EUR',
+  };
```
Yannick1712, line 50: Maybe set as private variable above constructor or as const like `TransactionBankTxTypeMapper`?

src/subdomains/generic/user/models/user/user.service.ts
```diff
@@ -240,15 +245,34 @@ export class UserService {
...
+    await this.userRepo.save(userToUpdate);
+
+    return { user: UserDtoMapper.mapUser(userData) };
+  }
+
+  async deactivateUser(id: number, address?: string): Promise<void> {
     const mainUser = await this.userRepo.findOne({ where: { id }, relations: { userData: { users: true } } });
     if (!mainUser) throw new NotFoundException('User not found');
     if (mainUser.userData.isBlocked || mainUser.userData.isDeactivated)
       throw new BadRequestException('User Account already deactivated');
```
Yannick1712, line 268: We don't need the mainUser anymore I would say. I would change `id` to `userDataId` and instead of the `mainUser` I would get the current userData with the `userDataId` (with `users` relation) directly at this point

src/subdomains/generic/user/models/user/user.service.ts
```diff
@@ -240,15 +245,34 @@ export class UserService {
...
-    if (mainUser.isBlockedOrDeactivated) throw new BadRequestException('User already deactivated');
 
-    if (!allUser) {
-      await this.userRepo.update(...mainUser.deactivateUser('Manual user deactivation'));
+    if (address) {
+      const user = mainUser.userData.users.find((u) => u.address === address);
+      if (!user) throw new NotFoundException('User not found');
+      if (user.status === UserStatus.BLOCKED) throw new BadRequestException('User already deactivated');
```
Yannick1712, line 273: Use `user.isBlockedOrDeactivated`

src/subdomains/generic/user/models/user/user.service.ts
```diff
@@ -240,15 +245,34 @@ export class UserService {
...
+
+      await this.userRepo.update(...user.deactivateUser('Manual user account block'));
```
Yannick1712, line 275: Change to `Manual user deactivation` from dev changes instead of `Manual user account block`

src/shared/models/fiat/fiat.service.ts
```diff
@@ -1,9 +1,19 @@
...
 @Injectable()
 export class FiatService {
+  private ipCountryToCurrency: { [key: string]: string } = {
+    DE: 'EUR',
+    AT: 'EUR',
+    CH: 'CHF',
+    LI: 'CHF',
+    IT: 'EUR',
+    FR: 'EUR',
+  };
```
Yannick1712, line 15: Maybe it makes sense to have a `ipCountryDefaultList` which then can be used for both `ipCountryToCurrency` and `ipCountryToLanguage` 🤔  With that we could define the key with the new `IpCountryDefaultList` 
@davidleomay wdym?

lapatric: Maybe also something like this in `geo-location.service.ts`? 🤔

```ts
  async getDefaultLanguage(ip: string): Promise<string> {
    const geoLocation = GeoIp.lookup(ip);
    const ipCountry = geoLocation?.country;
    switch (ipCountry) {
      case 'DE':
      case 'AT':
      case 'CH':
      case 'LI':
        return 'DE';
      case 'IT':
        return 'IT';
      case 'FR':
        return 'FR';
      default:
        return 'EN';
    }
  }

  async getDefaultCurrency(ip: string): Promise<string> {
    const geoLocation = GeoIp.lookup(ip);
    const ipCountry = geoLocation?.country;
    const isEU = geoLocation?.eu === '1';
    switch (ipCountry) {
      case 'CH':
      case 'LI':
        return 'CHF';
      default:
        return isEU ? 'EUR' : 'USD';
    }
  }
  ```

Comment (lapatric) [reply src/shared/models/fiat/fiat.service.ts:15]: Actually it doesn't look like the functions in `geo-location.service.ts` need to be `async`?

File: src/subdomains/generic/user/models/user/user.service.ts [hunk 7]
```diff
@@ -240,21 +245,42 @@ export class UserService {
...
+      where: { id: userDataId },
+      relations: { users: { wallet: true } },
+    });
+    if (!userData) throw new NotFoundException('User not found');
+    if (userData.isBlocked || userData.isDeactivated) throw new BadRequestException('User Account already deactivated');
+
+    if (address) {
+      const user = userData.users.find((u) => u.address === address);
+      if (!user) throw new NotFoundException('User not found');
+      if (user.isBlockedOrDeactivated) throw new BadRequestException('User already deactivated');
```
davidleomay, line 275: `'Address not found'`
`'Address already deactivated'`

src/subdomains/generic/user/models/user/user.service.ts
```diff
@@ -240,21 +245,42 @@ export class UserService {
...
+    if (!userData) throw new NotFoundException('User not found');
+
+    const userToUpdate = userData.users.find((u) => u.address === address);
+    if (!userToUpdate) throw new NotFoundException('Address not found');
+
+    userToUpdate.label = dto.label;
+    await this.userRepo.save(userToUpdate);
+
+    return { user: UserDtoMapper.mapUser(userData) };
+  }
```
davidleomay, line 262: Why not return the user directly?
`return UserDtoMapper.mapUser(userData);`

davidleomay (approved):
- [ ] Wait for [PR 1484](https://github.com/DFXswiss/api/pull/1484) to be merged (has some related changes)

davidleomay (approved):

lapatric (commented):
Merge conflict was in `updateAddress` and `deactivateUser` in `user.service.ts`

src/subdomains/generic/user/models/user/user.service.ts
```diff
@@ -240,21 +245,44 @@ export class UserService {
...
+    userToUpdate.label = dto.label;
+    await this.userRepo.save(userToUpdate);
+
+    return UserDtoMapper.mapUser(userData);
+  }
+
+  async deactivateUser(userDataId: number, address?: string): Promise<void> {
+    const userData = await this.userDataRepo.findOne({
+      where: { id: userDataId },
+      relations: { users: { wallet: true } },
```
Yannick1712, line 267: Relation `kycSteps` missing

src/subdomains/generic/user/models/user/user.service.ts
```diff
@@ -240,21 +245,44 @@ export class UserService {
...
+    });
+    if (!userData) throw new NotFoundException('User not found');
+    if (userData.isBlocked || userData.isDeactivated) throw new BadRequestException('User Account already deactivated');
```
Yannick1712, line 270: Could also use `userData.isBlockedOrDeactivated`

src/subdomains/generic/user/models/user/user.service.ts
```diff
@@ -240,21 +245,44 @@ export class UserService {
...
+
+    return UserDtoMapper.mapUser(userData);
+  }
+
+  async deactivateUser(userDataId: number, address?: string): Promise<void> {
+    const userData = await this.userDataRepo.findOne({
+      where: { id: userDataId },
+      relations: { users: { wallet: true } },
+    });
+    if (!userData) throw new NotFoundException('User not found');
```
Yannick1712, line 269: I would change message to `User Account not found`

src/subdomains/generic/user/models/user/user.controller.ts
```diff
@@ -228,6 +231,34 @@ export class UserV2Controller {
...
+    return this.userService.deactivateUser(jwt.account, address);
+  }
+
+  @Delete()
+  @ApiBearerAuth()
+  @UseGuards(AuthGuard(), new RoleGuard(UserRole.ACCOUNT))
+  @ApiOkResponse()
+  async deleteAccount(@GetJwt() jwt: JwtPayload): Promise<void> {
+    return this.userService.deactivateUser(jwt.account);
+  }
```
Yannick1712, line 260: Could we change the Guard in the deprecated endpoint above to `Account`?
In the moment this endpoint would not be reached because there is the deprecated Delete endpoint above. Else we should add a `TODO: remove` on the above Delete endpoint or?

davidleomay: This is in the v2 controller, so there should be no clashes.

src/subdomains/generic/user/models/user/user.controller.ts
```diff
@@ -228,6 +231,34 @@ export class UserV2Controller {
...
+
+  @Delete('addresses/:address')
+  @ApiBearerAuth()
+  @UseGuards(AuthGuard(), new RoleGuard(UserRole.ACCOUNT))
+  @ApiOkResponse()
+  async deleteAddress(@GetJwt() jwt: JwtPayload, @Param('address') address: string): Promise<void> {
+    return this.userService.deactivateUser(jwt.account, address);
+  }
+
+  @Delete()
```
Yannick1712, line 254: Maybe add `account` like the deprecated endpoint above?

src/subdomains/generic/user/models/user/user.service.ts
```diff
@@ -240,21 +245,44 @@ export class UserService {
...
-    if (!mainUser) throw new NotFoundException('User not found');
-    if (mainUser.userData.isBlockedOrDeactivated) throw new BadRequestException('User Account already deactivated');
-    if (mainUser.isBlockedOrDeactivated) throw new BadRequestException('User already deactivated');
+    if (!userData) throw new NotFoundException('User not found');
+
+    const userToUpdate = userData.users.find((u) => u.address === address);
+    if (!userToUpdate) throw new NotFoundException('Address not found');
+
+    userToUpdate.label = dto.label;
+    await this.userRepo.save(userToUpdate);
```
Yannick1712, line 259: Would be better to use the `update` method.
I would make an `setLabel` factory method in the user entity. Similar to `activateUser`. Just name it setLabel and return `UpdateResult<User>`, same as in `activateUser` and then you can easy use `await this.userRepo.update(...userToUpdate.setLabel(dto.label))`

src/subdomains/generic/user/models/user/user.service.ts
```diff
@@ -240,21 +245,44 @@ export class UserService {
...
+      await this.userRepo.update(...user.deactivateUser('Manual user deactivation'));
       return;
     }
 
-    await this.userDataService.deactivateUserData(mainUser.userData);
+    await this.userDataService.deactivateUserData(userData);
+
+    for (const user of userData.users) {
+      await this.userRepo.update(...user.deactivateUser('Manual user account deactivation'));
+    }
```
Yannick1712, line 285: Remove these lines here. We only deactivate the userData without the users from now on

Yannick1712 (approved):
Add migration

migration/1721730420883-AddressLabel.js
```diff
@@ -0,0 +1,15 @@
...
+    name = 'AddressLabel1721730420883'
+
+    async up(queryRunner) {
+        await queryRunner.query(`ALTER TABLE "wallet" ADD "displayName" nvarchar(256)`);
+        await queryRunner.query(`ALTER TABLE "user" ADD "label" nvarchar(256)`);
+    }
+
+    async down(queryRunner) {
+        await queryRunner.query(`ALTER TABLE "user" DROP COLUMN "label"`);
+        await queryRunner.query(`ALTER TABLE "wallet" DROP COLUMN "displayName"`);
```
Yannick1712, line 13: Add `"dbo"`

For example: 
`ALTER TABLE "dbo"."wallet" ADD "displayName" nvarchar(256)`
`ALTER TABLE "dbo"."user" ADD "label" nvarchar(256)`

Same for the down part

Yannick1712 (approved):

migration/1721730420883-AddressLabel.js
davidleomay: Should work without `dbo` IMO

davidleomay (approved):

PR #1514, opened 2024-08-04
[DEV-2932] NationalityStep refactoring

Yannick1712:
#### Pre-Release

- [ ] Check migrations
  - Login/logout
  - Buy/sell payment request
  - KYC page

#### Post-Release

- Test basic user operations

### DIFF
--- a/src/subdomains/generic/kyc/entities/kyc-step.entity.ts
+++ b/src/subdomains/generic/kyc/entities/kyc-step.entity.ts
@@ -9,10 +9,6 @@ import { StepLog } from './step-log.entity';
 
 export type KycStepResult = string | object;
 
-export interface KycStepNationalityData {
-  nationality: { id: number; symbol: string };
-}
-
 @Entity()
 @Index((s: KycStep) => [s.userData, s.name, s.type, s.sequenceNumber], { unique: true })
 export class KycStep extends IEntity {
--- a/src/subdomains/generic/kyc/entities/kyc-step.entity.ts
+++ b/src/subdomains/generic/kyc/entities/kyc-step.entity.ts
@@ -9,6 +9,10 @@ import { StepLog } from './step-log.entity';
 
 export type KycStepResult = string | object;
 
+export interface KycStepNationalityData {
+  nationality: { id: number; symbol: string };
+}
+
 @Entity()
 @Index((s: KycStep) => [s.userData, s.name, s.type, s.sequenceNumber], { unique: true })
 export class KycStep extends IEntity {
--- a/src/subdomains/generic/kyc/services/kyc.service.ts
+++ b/src/subdomains/generic/kyc/services/kyc.service.ts
@@ -271,7 +271,6 @@ export class KycService {
     stepId: number,
     data: Partial<UserData>,
     requiresInternalReview: boolean,
-    updateUserData: boolean,
   ): Promise<KycResultDto> {
     let user = await this.getUser(kycHash);
     const kycStep = user.getPendingStepOrThrow(stepId);
@@ -279,11 +278,9 @@ export class KycService {
     if (data.nationality) {
       const nationality = await this.countryService.getCountry(data.nationality.id);
       if (!nationality) throw new BadRequestException('Nationality not found');
-
-      data.nationality.symbol = nationality.symbol;
+    } else {
+      user = await this.userDataService.updateUserDataInternal(user, data);
     }
-
-    if (updateUserData) user = await this.userDataService.updateUserDataInternal(user, data);
 
     user = requiresInternalReview ? user.internalReviewStep(kycStep, data) : user.completeStep(kycStep, data);
     await this.createStepLog(user, kycStep);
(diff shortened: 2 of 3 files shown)


src/subdomains/generic/kyc/services/kyc.service.ts
```diff
@@ -266,16 +266,24 @@ export class KycService {
     return KycStepMapper.toKycResult(kycStep);
   }
 
-  async updateUserData(
+  async updateKycStep(
     kycHash: string,
     stepId: number,
     data: Partial<UserData>,
     requiresInternalReview: boolean,
+    updateUserData: boolean,
```
Yannick1712, line 274: @davidleomay makes one more param sense or should we just make two methods?

src/subdomains/generic/kyc/services/kyc.service.ts
```diff
@@ -266,16 +266,24 @@ export class KycService {
...
   ): Promise<KycResultDto> {
     let user = await this.getUser(kycHash);
     const kycStep = user.getPendingStepOrThrow(stepId);
 
-    user = await this.userDataService.updateUserDataInternal(user, data);
+    if (data.nationality) {
+      const nationality = await this.countryService.getCountry(data.nationality.id);
+      if (!nationality) throw new BadRequestException('Nationality not found');
+
+      data.nationality.symbol = nationality.symbol;
```
Yannick1712, line 283: @davidleomay wdyt, is it better to have nationality: {id: number, symbol: string} as kycStepResult or just nationality: string with the symbol instead of the id? 🤔 If we decide to split into two methods we could do that or do we need the countryId in the stepResult, too?

src/subdomains/generic/kyc/services/kyc.service.ts
Yannick1712, line 1: @davidleomay In the updateFileData method we only save the file url as string. For every other kycStepResult we have done a json as result. Maybe we should consider something like url: string as the fileResult? 🤔

src/subdomains/generic/kyc/services/kyc.service.ts
davidleomay: I think 2 methods would be cleaner 🤔

src/subdomains/generic/kyc/services/kyc.service.ts
davidleomay: I think it's ok to just store the data as it comes, the ID should be enough, don't you think?

src/subdomains/generic/kyc/services/kyc.service.ts
davidleomay: Is probably a bit late now, there are already a bunch of steps with URL results.

src/subdomains/generic/kyc/services/kyc.service.ts
Yannick1712: Yeah thought that too, it's a bit strange to build it in 2 methods because it's only 1 line that differs (the updateUserData line) and above and below the code is the same in both methods 😅

Yannick1712: I have pushed another solution you can look at it maybe it's acceptable too 👍

src/subdomains/generic/kyc/entities/kyc-step.entity.ts
```diff
@@ -9,6 +9,10 @@ import { StepLog } from './step-log.entity';
 
 export type KycStepResult = string | object;
 
+export interface KycStepNationalityData {
+  nationality: { id: number; symbol: string };
+}
```
davidleomay, line 14: Remove this and use IEntity instead?

src/subdomains/generic/kyc/services/kyc.service.ts
```diff
@@ -627,8 +636,11 @@ export class KycService {
     return userData;
   }
 
-  private getIdentCheckErrors(entity: KycStep, result: IdentResultDto): IdentCheckError[] {
+  private getIdentCheckErrors(entity: KycStep, result: IdentResultDto, nationality?: Country): IdentCheckError[] {
     const errors = [];
+    const nationalityStepResult = entity.userData.kycSteps
+      .find((k) => k.name === KycStepName.NATIONALITY_DATA && k.status === KycStepStatus.COMPLETED)
+      .getResult<KycStepNationalityData>();
```
davidleomay, line 643: `entity.userData.getStepsWith(KycStepName.NATIONALITY_DATA).find((s) => s.isCompleted)`

src/subdomains/generic/kyc/services/kyc.service.ts
```diff
@@ -641,8 +653,7 @@ export class KycService {
     )
       errors.push(IdentCheckError.LAST_NAME_NOT_MATCHING);
 
-    if (entity.userData.nationality?.symbol !== result.userdata.nationality?.value)
-      errors.push(IdentCheckError.NATIONALITY_NOT_MATCHING);
+    if (nationalityStepResult.nationality.id !== nationality?.id) errors.push(IdentCheckError.NATIONALITY_NOT_MATCHING);
```
davidleomay, line 656: `nationalityStepResult` might be undefined?

davidleomay (approved):

PR #1516, opened 2024-08-05
[DEV-2936] new amlReasons userBlocked

Yannick1712:
#### Pre-Release

- [ ] Check migrations
  - Login/logout
  - Buy/sell payment request
  - KYC page

#### Post-Release

- Test basic user operations


src/integration/sift/dto/sift.dto.ts
```diff
@@ -1008,6 +1008,9 @@ export const SiftAmlDeclineMap: { [method in AmlReason]: DeclineCategory } = {
   [AmlReason.CHF_ABROAD_TX]: DeclineCategory.INVALID,
   [AmlReason.ASSET_KYC_NEEDED]: DeclineCategory.OTHER,
   [AmlReason.CARD_NAME_MISMATCH]: DeclineCategory.OTHER,
+  [AmlReason.USER_BLOCKED]: DeclineCategory.OTHER,
+  [AmlReason.USER_DATA_BLOCKED]: DeclineCategory.OTHER,
```
davidleomay, line 1012: User blocked is probably `RISKY`?

src/subdomains/generic/user/models/user/user.entity.ts
```diff
@@ -179,4 +179,12 @@ export class User extends IEntity {
   get isBlockedOrDeactivated(): boolean {
     return [UserStatus.BLOCKED, UserStatus.DEACTIVATED].includes(this.status);
   }
```
davidleomay, line 181: Use `isBlocked` and `isDeactivated` from below?

src/subdomains/supporting/payment/dto/transaction.dto.ts
```diff
@@ -46,6 +46,7 @@ export enum TransactionReason {
   CHF_ABROAD_NOT_ALLOWED = 'ChfAbroadNotAllowed',
   ASSET_KYC_NEEDED = 'AssetKycNeeded',
   CARD_NAME_MISMATCH = 'CardNameMismatch',
+  USER_DEACTIVATED = 'UserDeactivated',
 }
```
davidleomay, line 50: Update services

src/integration/sift/dto/sift.dto.ts
Yannick1712: Hm, could also be that we blocked him because his wallet got hacked for example 🤔 don't know if that RISKY Option would lead to problems for this users, but eah we could do that 😅

davidleomay (approved):

src/subdomains/supporting/payment/dto/transaction.dto.ts
davidleomay, line 50: Maybe we should call it USER_DELETED, because we externally speak of deletion, not deactivation?

Yannick1712: Yeah would be make sense, I clarify that

Yannick1712:
- [ ] update already deactivated user

davidleomay (approved):

PR #1518, opened 2024-08-07
[DEV-2904] Mail confirmation

lapatric:
#### Pre-Release

- [ ] Check migrations
  - Login/logout
  - Buy/sell payment request
  - KYC page

#### Post-Release

- Test basic user operations


src/subdomains/generic/user/models/account-merge/account-merge.service.ts
```diff
@@ -72,7 +72,10 @@ export class AccountMergeService {
   }
 
   async executeMerge(code: string): Promise<AccountMerge> {
-    const request = await this.accountMergeRepo.findOne({ where: { code }, relations: { master: true, slave: true } });
+    const request = await this.accountMergeRepo.findOne({
+      where: { code },
+      relations: { master: { users: true }, slave: { users: true } },
+    });
```
davidleomay, line 78: Can be reverted?

src/subdomains/generic/user/models/account-merge/account-merge.service.ts
```diff
@@ -84,6 +87,6 @@ export class AccountMergeService {
   }
 
   private buildConfirmationUrl(code: string): string {
-    return `${Config.url()}/auth/mail/confirm?code=${code}`;
+    return `${Config.frontend.services}/kyc?merge-code=${code}`;
   }
```
davidleomay, line 91: Are you sure you want to do it on the KYC screen? Maybe a separate screen is easier?

davidleomay: Adapt Mail with new link

src/subdomains/generic/user/models/auth/auth.controller.ts
```diff
@@ -61,10 +69,25 @@ export class AuthController {
...
+    const newUser = await this.userRepo.findOne({
+      where: { userData: { id: master.id }, address: jwt.address },
+      relations: { userData: true, wallet: true },
+    });
+
+    return {
+      redirectUrl: master.kycUrl,
+      accessToken: this.authService.generateUserToken(newUser, ip),
+    };
   }
```
davidleomay, line 91: Ok, I see some issues here:
- The user will click a link he received by mail. Therefore he might not be logged in in the Services
   - We cannot have an auth guard
   - We can only return an access token, if the user came with one
   - Maybe we should remove the `redirectUrl` in the response and just return the master's KYC code and the JWT, if the user was logged in
- The User might have an "account-only" account, which means he has no user
   - In this case (there is no `user` in the JWT) use `generateAccountToken`

davidleomay (approved):

davidleomay (approved):

davidleomay (approved):

PR #1533, opened 2024-08-16
[DEV-2933] payment quote webhook

bernd2022:
#### Pre-Release

- [ ] Check migrations
  - Login/logout
  - Buy/sell payment request
  - KYC page

#### Post-Release

- Test basic user operations

### DIFF
--- a/src/shared/utils/queue-handler.ts
+++ b/src/shared/utils/queue-handler.ts
@@ -72,9 +72,8 @@ export class QueueHandler {
     while (this.isRunning) {
       try {
         if (this.queue.length > 0) {
-          this.workParallel
-            ? void this.queue.shift().doWork(this.itemTimeout)
-            : await this.queue.shift().doWork(this.itemTimeout);
+          const work = this.queue.shift().doWork(this.itemTimeout);
+          if (!this.workParallel) await work;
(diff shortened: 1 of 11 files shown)


src/subdomains/core/payment-link/services/payment-webhook.service.ts
davidleomay, line 1: Hm, the bad thing about the `QueueHandler` is that it has a single, blocking runner. So a failing webhook will block all others. We probably need to adapt the queue handler, so it can be configured to handle items in parallel.

src/shared/utils/queue-handler.ts
```diff
@@ -68,7 +72,9 @@ export class QueueHandler {
     while (this.isRunning) {
       try {
         if (this.queue.length > 0) {
-          await this.queue.shift().doWork(this.itemTimeout);
+          this.workParallel
+            ? void this.queue.shift().doWork(this.itemTimeout)
+            : await this.queue.shift().doWork(this.itemTimeout);
```
davidleomay, line 77:
```ts
const work = this.queue.shift().doWork(this.itemTimeout);
if (!this.workParallel) await work;
```

PR #1540, opened 2024-08-19
[DEV-2942] VideoIdentNeeded

Yannick1712:
#### Pre-Release

- [ ] Check migrations
  - Login/logout
  - Buy/sell payment request
  - KYC page

#### Post-Release

- Test basic user operations

### DIFF
--- a/src/subdomains/generic/kyc/services/kyc-admin.service.ts
+++ b/src/subdomains/generic/kyc/services/kyc-admin.service.ts
@@ -1,5 +1,6 @@
 import { Inject, Injectable, NotFoundException, forwardRef } from '@nestjs/common';
 import { CountryService } from 'src/shared/models/country/country.service';
+import { DfxLogger } from 'src/shared/services/dfx-logger';
 import { DisabledProcess, Process } from 'src/shared/services/process.service';
 import { BankDataType } from '../../user/models/bank-data/bank-data.entity';
 import { BankDataService } from '../../user/models/bank-data/bank-data.service';
@@ -15,6 +16,8 @@ import { KycService } from './kyc.service';
 
 @Injectable()
 export class KycAdminService {
+  private readonly logger = new DfxLogger(KycAdminService);
+
   constructor(
     private readonly kycStepRepo: KycStepRepository,
     private readonly webhookService: WebhookService,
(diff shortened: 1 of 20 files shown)


src/subdomains/supporting/payment/dto/transaction.dto.ts
```diff
@@ -47,6 +47,7 @@ export enum TransactionReason {
   ASSET_KYC_NEEDED = 'AssetKycNeeded',
   CARD_NAME_MISMATCH = 'CardNameMismatch',
   USER_DELETED = 'UserDeleted',
+  VIDEO_IDENT_NEEDED = 'VideoIdentNeeded',
 }
```
davidleomay, line 51: Update Services

src/subdomains/generic/kyc/services/kyc.service.ts
```diff
@@ -514,7 +514,8 @@ export class KycService {
         return {
           nextStep: {
             name: nextStep,
-            type: await this.userDataService.getIdentMethod(user),
+            type:
+              lastTry.type === KycStepType.VIDEO ? KycStepType.VIDEO : await this.userDataService.getIdentMethod(user),
```
davidleomay, line 518: `lastTry` might be null

src/subdomains/generic/kyc/services/kyc-admin.service.ts
```diff
@@ -62,6 +62,10 @@ export class KycAdminService {
     }
   }
 
+  async triggerVideoIdentInternal(userData: UserData): Promise<void> {
+    await this.kycService.initiateStep(userData, KycStepName.IDENT, KycStepType.VIDEO);
+  }
```
davidleomay, line 67: Maybe we should check if there is already a pending video ident?

Yannick1712: Yeah thought that too, already 👍

Yannick1712: Ah I see we got that already, don't need to change anything I would say 👍 check `identService`
Maybe it would make sense to add a try&catch and log the error in case 👍 will add that 👍

src/subdomains/generic/kyc/services/kyc.service.ts
```diff
@@ -530,7 +545,7 @@ export class KycService {
     }
   }
 
-  private async initiateStep(
+  async initiateStep(
```
davidleomay, line 548: Revert

davidleomay (approved):

PR #1578, opened 2024-09-03
[DEV-3055] Route table

Yannick1712:
#### Pre-Release

- [ ] Check migrations
  - Login/logout
  - Buy/sell payment request
  - KYC page

#### Post-Release

- Test basic user operations


Yannick1712:
- [ ] add migration

src/subdomains/core/route/route.controller.ts
davidleomay, line 1: There is already a `RouteController`, but it's a bit misplaced 🤔

src/subdomains/core/route/route.controller.ts
```diff
@@ -0,0 +1,32 @@
...
+export class RouteController {
+  constructor(private readonly routeService: RouteService) {}
+
+  @Post()
+  @ApiBearerAuth()
+  @UseGuards(AuthGuard(), new RoleGuard(UserRole.ADMIN))
+  @ApiExcludeEndpoint()
+  async createRoute(@Body() dto: CreateRouteDto): Promise<Route> {
+    return this.routeService.createRoute(dto);
+  }
```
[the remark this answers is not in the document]
Yannick1712: Yeah if label is nullable true thats the best way 👍

src/subdomains/core/route/route.controller.ts
Yannick1712: Hm yeah 🤔 what should we do about it? Who is using the old endpoint? If e.g. only services uses it, we could rename it to depositRoute controller 👍

davidleomay: Can't we move the existing code into the new controller?

Yannick1712: Haha ah yeah true 👍

src/subdomains/core/route/route.entity.ts
```diff
@@ -0,0 +1,37 @@
...
+import { User } from 'src/subdomains/generic/user/models/user/user.entity';
+import { Column, Entity, OneToOne } from 'typeorm';
+import { Buy } from '../buy-crypto/routes/buy/buy.entity';
+import { Swap } from '../buy-crypto/routes/swap/swap.entity';
+import { Sell } from '../sell-crypto/route/sell.entity';
+
+@Entity()
+export class Route extends IEntity {
+  @Column({ length: 256, unique: true, nullable: true })
+  label: string;
```
davidleomay, line 12: I think this unique index is not helpful/required?

src/subdomains/core/route/route.service.ts
```diff
@@ -0,0 +1,93 @@
...
+    const entity = await this.routeRepo.findOne({
+      where: { id },
+      relations: { buy: true, sell: true, swap: true },
+    });
+    if (!entity) throw new NotFoundException('Route not found');
+
+    const update = this.routeRepo.create(dto);
+
+    if (update.label && (await this.routeRepo.existsBy({ label: update.label })))
+      throw new BadRequestException('Label already in use');
```
davidleomay, line 83: Same here

src/subdomains/core/route/route.service.ts
```diff
@@ -0,0 +1,93 @@
...
+
+    if (dto.sell) {
+      entity.sell = await this.sellService.get(undefined, dto.sell.id);
+      if (!entity.sell) throw new NotFoundException('Sell not found');
+    }
+
+    if (dto.swap) {
+      entity.swap = await this.swapService.get(undefined, dto.swap.id);
+      if (!entity.swap) throw new NotFoundException('Swap not found');
+    }
```
davidleomay, line 68: I think these checks can be removed. The method is only used internally with an existing route.

src/subdomains/core/route/route.service.ts
davidleomay, line 1: I think we should add some code create a route directly when creating a buy/sell/swap. Or do you want to add this later?

src/subdomains/core/route/route.entity.ts
Yannick1712: That's what the ticket says or? 🤔 
Robin said in the ticket, the label col has to be unique on the complete table

src/subdomains/core/route/route.service.ts
Yannick1712: Yes right, thats a second step. First sync all existing routes

src/subdomains/core/route/route.entity.ts
davidleomay: Ah, true, you're right

src/subdomains/core/route/route.service.ts
```diff
@@ -0,0 +1,82 @@
...
+      }
+    }
+  }
+
+  async createRoute(dto: CreateRouteDto): Promise<Route> {
+    const entity = this.routeRepo.create(dto);
+
+    if (dto.buy) entity.buy = await this.buyService.get(undefined, dto.buy.id);
+    if (dto.sell) entity.sell = await this.sellService.get(undefined, dto.sell.id);
+    if (dto.swap) entity.swap = await this.swapService.get(undefined, dto.swap.id);
```
davidleomay, line 57: Remove

davidleomay (approved):

PR #1664, opened 2024-10-03
[DEV-3063] bankAccount bankData sync

Yannick1712:
#### Pre-Release

- [ ] Check migrations
  - Login/logout
  - Buy/sell payment request
  - KYC page

#### Post-Release

- Test basic user operations

### DIFF
--- a/src/subdomains/generic/user/models/bank-data/bank-data.entity.ts
+++ b/src/subdomains/generic/user/models/bank-data/bank-data.entity.ts
@@ -1,4 +1,5 @@
 import { IEntity, UpdateResult } from 'src/shared/models/entity';
+import { Fiat } from 'src/shared/models/fiat/fiat.entity';
 import { UserData } from 'src/subdomains/generic/user/models/user-data/user-data.entity';
 import { Column, Entity, Index, ManyToOne } from 'typeorm';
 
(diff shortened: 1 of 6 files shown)


Yannick1712:
- [ ] add migration

src/subdomains/generic/user/models/bank-data/bank-data.entity.ts
```diff
@@ -20,6 +21,10 @@ export enum BankDataVerificationError {
 }
 
 @Entity()
+@Index((bankData: BankData) => [bankData.iban, bankData.userData], {
+  unique: true,
+  where: `type = '${BankDataType.USER}'`,
+})
```
davidleomay, line 27: Does this interfere with current data?

src/subdomains/supporting/bank/bank-account/bank-account.entity.ts
```diff
@@ -36,6 +36,9 @@ export class BankAccount extends IEntity implements BankAccountInfos {
   @Column({ length: 256 })
   iban: string;
 
+  @Column({ nullable: true })
+  synced: boolean;
```
davidleomay, line 40: `default: false`?

src/subdomains/generic/user/models/bank-data/bank-data.entity.ts
Yannick1712: Jap that should not be a problem 👍  I checked it manually in prd DB. Looks good 👍

src/subdomains/supporting/bank/bank-account/bank-account.entity.ts
Yannick1712: I think, we don't need it. The column is only temporary for the sync 👍

davidleomay (approved):
- [x] Add migration

PR #1671, opened 2024-10-08
[DEV-3217] financeLog pending Amounts

Yannick1712:
#### Pre-Release

- [ ] Check migrations
  - Login/logout
  - Buy/sell payment request
  - KYC page

#### Post-Release

- Test basic user operations

### DIFF
--- a/src/integration/exchange/entities/exchange-tx.entity.ts
+++ b/src/integration/exchange/entities/exchange-tx.entity.ts
@@ -1,4 +1,7 @@
+import { Asset } from 'src/shared/models/asset/asset.entity';
 import { IEntity } from 'src/shared/models/entity';
+import { BankService } from 'src/subdomains/supporting/bank/bank/bank.service';
+import { BankExchangeType } from 'src/subdomains/supporting/log/log-job.service';
 import { Column, Entity, Index } from 'typeorm';
 import { ExchangeName } from '../enums/exchange.enum';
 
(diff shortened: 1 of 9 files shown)


src/subdomains/supporting/log/log-job.service.ts
```diff
@@ -87,6 +93,22 @@ export class LogJobService {
...
+      BankTxType.KRAKEN,
+    );
+    const pendingKrakenTx = await this.exchangeTxService
+      .getPendingExchangeTx(
+        ExchangeTxType.WITHDRAWAL,
+        ExchangeName.KRAKEN,
+        'Maerki Baumann & Co. AG',
+        'Bank Frick (SEPA)',
+      )
+      .then((p) => p.filter((p) => p.currency === 'EUR'));
```
davidleomay, line 111: Maybe this methods should be renamed to `getRecent...`, they are not pending 😉

src/integration/exchange/entities/exchange-tx.entity.ts
davidleomay, line 1: Why these changes? 🤔

davidleomay (commented):
- [x] Get last 7 days of receiver and 14 days of sender transactions (instead of fixed IDs)

src/integration/exchange/entities/exchange-tx.entity.ts
Yannick1712: It's not working with Entity Methods these way. With entity methods you would have to implement these on every position you used the dto 😅

davidleomay: You can just use `export type ExchangeTxDto = Omit<ExchangeTx, keyof IEntity | 'pendingBankAmount'>;`

Yannick1712: Yeah sure, but that's a pain with every entity method we will add 🤔 Thought about this solution too but for me this solution isn't so nice, instead of just creating a real dto 🤔 but yeah I we can do this. 
Yeah with the extra dto class we don't need the `?`, because they are defined in the dto itself but clearly if we change that we have to add them in the entity again 👍

davidleomay (approved):

PR #1695, opened 2024-10-15
[DEV-2197] userData replace save

Yannick1712:
#### Pre-Release

- [ ] Check migrations
  - Login/logout
  - Buy/sell payment request
  - KYC page

#### Post-Release

- Test basic user operations


src/subdomains/generic/user/models/user-data/user-data.entity.ts
davidleomay, line 1: Remove `{ cascade: true }` on `kycSteps` property (might need a migration)

davidleomay (approved):

PR #1741, opened 2024-10-25
[DEV-3296] CT userData api key

Yannick1712:
#### Pre-Release

- [ ] Check migrations
  - Login/logout
  - Buy/sell payment request
  - KYC page

#### Post-Release

- Test basic user operations

### DIFF
--- a/src/subdomains/core/history/controllers/history.controller.ts
+++ b/src/subdomains/core/history/controllers/history.controller.ts
@@ -26,7 +26,6 @@ import { UserRole } from 'src/shared/auth/user-role.enum';
 import { ApiKeyService } from 'src/shared/services/api-key.service';
 import { Util } from 'src/shared/utils/util';
 import { UserDataService } from 'src/subdomains/generic/user/models/user-data/user-data.service';
-import { User } from 'src/subdomains/generic/user/models/user/user.entity';
 import { UserService } from 'src/subdomains/generic/user/models/user/user.service';
 import { TransactionDto } from '../../../supporting/payment/dto/transaction.dto';
 import { ExportFormat, HistoryQuery, HistoryQueryExportType, HistoryQueryUser } from '../dto/history-query.dto';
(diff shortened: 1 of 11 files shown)


Yannick1712:
-[ ] add migration

Yannick1712:
- [ ] update services

src/subdomains/generic/user/models/user-data/user-data.controller.ts
```diff
@@ -35,6 +39,31 @@ export class UserDataController {
...
+    return this.userDataService.deleteApiKey(jwt.user);
+  }
+
+  @Put('apiFilter/CT')
+  @ApiBearerAuth()
+  @UseGuards(AuthGuard(), new RoleGuard(UserRole.USER))
+  @ApiOkResponse({ type: String, isArray: true })
+  async updateApiFilter(@GetJwt() jwt: JwtPayload, @Query() filter: HistoryFilter): Promise<HistoryFilterKey[]> {
+    return this.userDataService.updateApiFilter(jwt.user, filter);
+  }
```
davidleomay, line 65: I would keep it in the UserController (we don't have anything public in the UserDataController), the user doesn't know about UserData

davidleomay: `UserRole.ACCOUNT`?

src/subdomains/core/history/controllers/history.controller.ts
```diff
@@ -68,14 +71,26 @@ export class HistoryController {
     @Headers('DFX-ACCESS-SIGN') sign: string,
     @Headers('DFX-ACCESS-TIMESTAMP') timestamp: string,
   ): Promise<CoinTrackingApiHistoryDto[]> {
-    const user = await this.userService.checkApiKey(key, sign, timestamp);
+    const user =
+      key.charAt(key.length - 1) === '0'
```
davidleomay, line 75: `.endsWith('0')`

src/subdomains/core/history/controllers/history.controller.ts
```diff
@@ -68,14 +71,26 @@ export class HistoryController {
...
+        ? await this.historyService.getJsonHistory(
+            { format: ExportFormat.JSON, userAddress: user.address, ...query },
+            ExportType.COIN_TRACKING,
+          )
+        : await Promise.all(
+            user.users.map(async (u) => {
+              return await this.historyService.getJsonHistory(
+                { format: ExportFormat.JSON, userAddress: u.address, ...query },
+                ExportType.COIN_TRACKING,
+              );
```
davidleomay, line 91: Linter warning

src/subdomains/core/history/controllers/history.controller.ts
```diff
@@ -68,14 +71,26 @@ export class HistoryController {
...
+            }),
+          ).then((t) => t.flat())
```
davidleomay, line 93: Can't we use the new transaction history thing instead of using the Promise.all (which is probably more expensive)?

src/subdomains/generic/user/models/user-data/user-data.controller.ts
Yannick1712: Ah ok good point so you would then say, we change it instantly in userController, from userService to userDataService methods?

src/subdomains/core/history/controllers/history.controller.ts
Yannick1712: I think I tried it, but will take a look in it again 🤔

src/subdomains/generic/user/models/user-data/user-data.controller.ts
Yannick1712: Or should we give the option to give the version number as parameter? 🤔

davidleomay: I think I would create the API key on user data, for all new keys. For existing ones, we would need to update/delete it on the user. WDYT?

Yannick1712: Jap would say the same 👍

davidleomay (approved):

PR #1760, opened 2024-11-06
[DEV-3096] extend financialAssetLog

Yannick1712:
#### Pre-Release

- [ ] Check migrations
  - Login/logout
  - Buy/sell payment request
  - KYC page

#### Post-Release

- Test basic user operations


src/subdomains/supporting/log/log-job.service.ts
```diff
@@ -167,12 +168,28 @@ export class LogJobService {
...
+    // deposit address balance
+    const paymentAssets = assets.filter((a) => a.paymentEnabled);
+    const paymentAssetMap = Util.groupBy<Asset, Blockchain>(paymentAssets, 'blockchain');
+
+    const depositBalances = await Promise.all(
+      Array.from(paymentAssetMap.entries()).map(async ([e, a]) => {
+        const client = this.evmRegistryService.getClient(e);
+        const balances = await this.getCustomBalances(client, a, [
+          EvmUtil.createWallet({ seed: Config.payment.evmSeed, index: 0 }).address,
+        ]).then((b) => b.flat());
```
davidleomay, line 192: Hm ok, it's a bit more complex here 🤔  we also have Lightning/BTC and Monero/XMR which are payment enabled. Lightning we can ignore, because there is no separate address. But for XMR there is one (see payment config) and we need to get the balance of this address.

src/subdomains/supporting/log/log-job.service.ts
```diff
@@ -283,8 +300,12 @@ export class LogJobService {
         .find((c) => c.blockchain === curr.blockchain)
         ?.balances?.reduce((sum, result) => sum + (result.contractAddress === curr.chainId ? result.balance : 0), 0);
 
+      const depositBalance = depositBalances
+        .find((c) => c.blockchain === curr.blockchain)
+        ?.balances?.reduce((sum, result) => sum + (result.contractAddress === curr.chainId ? result.balance : 0), 0);
```
davidleomay, line 305: Chain ID comparison does not work for Monero

src/subdomains/supporting/log/log-job.service.ts
Yannick1712: Hm ok so I would say we have to hardcoded it for monero and Lightning or? 🤔

src/subdomains/supporting/log/log-job.service.ts
```diff
@@ -167,12 +168,33 @@ export class LogJobService {
...
+    // deposit address balance
+    const paymentAssets = assets.filter((a) => a.paymentEnabled && a.blockchain !== Blockchain.LIGHTNING);
+    const paymentAssetMap = Util.groupBy<Asset, Blockchain>(paymentAssets, 'blockchain');
+
+    const depositBalances = await Promise.all(
+      Array.from(paymentAssetMap.entries()).map(async ([e, a]) => {
+        const client = this.evmRegistryService.getClient(e);
+        const balances = await this.getCustomBalances(client, a, [
+          e === Blockchain.MONERO
+            ? Config.payment.moneroAddress
```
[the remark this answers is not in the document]
Yannick1712: Yeah maybe the extending makes more sense I think 🤔 don't know how complicated it is

src/integration/blockchain/monero/monero-client.ts
```diff
@@ -57,6 +63,26 @@ export class MoneroClient {
...
+    throw new Error('Monero has no token');
+  }
+
+  async getTokenBalances(_: Asset[], __?: string): Promise<BlockchainTokenBalance[]> {
+    throw new Error('Monero has no token');
+  }
+
+  async isTxComplete(_: string, __?: number): Promise<boolean> {
+    throw new Error('Monero method not implemented.');
+  }
```
davidleomay, line 82: Implement it quickly?
```ts
const transaction = await this.getTransaction(txId);
return MoneroHelper.isTransactionComplete(transaction);
```
and use it everywhere, where `MoneroHelper.isTransactionComplete` is currently used.

davidleomay (approved):

PR #1766, opened 2024-11-09
[DEV-3304] Add valid and category column

Kolibri1990:
#### Pre-Release

- [ ] Check migrations
  - Login/logout
  - Buy/sell payment request
  - KYC page

#### Post-Release

- Test basic user operations

### DIFF
--- a/src/subdomains/supporting/log/log-job.service.ts
+++ b/src/subdomains/supporting/log/log-job.service.ts
@@ -114,8 +114,8 @@ export class LogJobService {
         //TODO reactivate
         //changes: changeLog,
       }),
-      valid: true,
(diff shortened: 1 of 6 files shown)


src/subdomains/supporting/log/log-job.service.ts
```diff
@@ -114,6 +114,8 @@ export class LogJobService {
         //TODO reactivate
         //changes: changeLog,
       }),
+      valid: true,
+      category: '',
     });
```
davidleomay, line 119:
```ts
      valid: null,
      category: null,
```

davidleomay (approved):

PR #1825, opened 2024-12-04
[DEV-3406] additional export data

lapatric:
#### Pre-Release

- [ ] Check migrations
  - Login/logout
  - Buy/sell payment request
  - KYC page

#### Post-Release

- Test basic user operations


src/config/config.ts
davidleomay, line 1: Looks like we can remove the reduce filter and just always take the latest one 😁

davidleomay (approved):

PR #1931, opened 2025-01-30
[DEV-3388] deposit mail warning addition

Yannick1712:
#### Pre-Release

- [ ] Check migrations
  - Login/logout
  - Buy/sell payment request
  - KYC page

#### Post-Release

- Test basic user operations

### DIFF
--- a/src/subdomains/core/buy-crypto/process/services/buy-crypto-notification.service.ts
+++ b/src/subdomains/core/buy-crypto/process/services/buy-crypto-notification.service.ts
@@ -1,5 +1,4 @@
 import { Injectable } from '@nestjs/common';
-import { SettingService } from 'src/shared/models/setting/setting.service';
 import { DfxLogger } from 'src/shared/services/dfx-logger';
 import { DisabledProcess, Process } from 'src/shared/services/process.service';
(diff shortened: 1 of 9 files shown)


Yannick1712:
- [ ] add warning wallets in settings

src/subdomains/core/buy-crypto/process/services/buy-crypto-notification.service.ts
```diff
@@ -86,6 +88,8 @@ export class BuyCryptoNotificationService {
     for (const entity of entities) {
       try {
         if (entity.userData.mail) {
+          const mailWarningWalletIds = await this.settingService.getObj<number[]>('mailWarningWalletIds', []);
+
```
davidleomay, line 92: Why not store it in the wallet table?

Yannick1712: As an additional column? Yeah could do that, I think 👍 
I thought it would not be so much wallets, but maybe there getting more which need this warning 😁

davidleomay: Yes, I would prefer this

davidleomay (approved):
- [x] Add migration

PR #1932, opened 2025-01-30
[DEV-3583] Check connection pool

bernd2022:
#### Pre-Release

- [ ] Check migrations
  - Login/logout
  - Buy/sell payment request
  - KYC page

#### Post-Release

- Test basic user operations

### DIFF
--- a/src/subdomains/generic/gs/gs.service.ts
+++ b/src/subdomains/generic/gs/gs.service.ts
@@ -1,4 +1,5 @@
 import { BadRequestException, Injectable, NotFoundException } from '@nestjs/common';
+import { Config } from 'src/config/config';
 import { DfxLogger } from 'src/shared/services/dfx-logger';
 import { Util } from 'src/shared/utils/util';
 import { BuyCryptoService } from 'src/subdomains/core/buy-crypto/process/services/buy-crypto.service';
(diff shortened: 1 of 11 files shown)


src/subdomains/generic/gs/gs.service.ts
```diff
@@ -408,4 +434,27 @@ export class GsService {
...
+    // Info, if there is a pending connection
+    if (this.dbConnectionPool.pending > 0) {
+      this.logger.info(
+        `ConnectionPool with pending connections: S${this.dbConnectionPool.size}/A${this.dbConnectionPool.available}/P${this.dbConnectionPool.pending}/B${this.dbConnectionPool.borrowed}/${runtime}ms/${remark}`,
+      );
+    }
+
+    // Warning, if there are all connections in use
+    const dbOptions = Config.database as SqlServerConnectionOptions;
+    const dbMaxPoolConnections = dbOptions.pool?.max ?? 10;
```
Yannick1712, line 449: remove `?? 10` already defined in config

Yannick1712, line 449: Do we need `?`, is pool optional? 🤔

src/subdomains/generic/gs/gs.service.ts
```diff
@@ -408,4 +434,27 @@ export class GsService {
...
+
+    if (
+      dbMaxPoolConnections === this.dbConnectionPool.size &&
+      dbMaxPoolConnections === this.dbConnectionPool.borrowed
```
Yannick1712, line 453: Is it possible to get a dbConnectionPool.size with more than the dbMaxPoolConnections or is it limited?

src/subdomains/generic/gs/gs.service.ts
bernd2022: Yes, it is optional. Refer to the "SqlServerConnectionOptions", its a "readonly pool?". Almost all of the parameters of the "SqlServerConnectionOptions" are optional.

src/subdomains/generic/gs/gs.service.ts
bernd2022: It is limited to the maximum given in the options. The default max. is 10 if no other value is provided.

src/subdomains/generic/gs/gs.service.ts
davidleomay, line 1: Wouldn't it be better to just add a cron job for checking the connection pool periodically, maybe every second?

davidleomay (approved):

PR #1944, opened 2025-02-06
[DEV-3583] DFX cron decorator and service

bernd2022:
#### Pre-Release

- [ ] Check migrations
  - Login/logout
  - Buy/sell payment request
  - KYC page

#### Post-Release

- Test basic user operations


src/shared/utils/util.ts
```diff
@@ -406,6 +409,36 @@ export class Util {
...
+      case CronExpression.EVERY_10_MINUTES:
+        return Util.delay(random * 30); // 0 .. 30 sec
+
+      case CustomCronExpression.EVERY_15_MINUTES:
+        return Util.delay(random * 30); // 0 .. 30 sec
+
+      case CronExpression.EVERY_HOUR:
+        return Util.delay(random * 60); // 0 .. 60 sec
+    }
+  }
```
davidleomay, line 440: I would move this as a private method to the DfxCronService

src/shared/services/dfx-cron.service.ts
```diff
@@ -0,0 +1,72 @@
...
+  private addCronJob(data: CronJobData) {
+    const lock = LockClass.create(data.params.timeout ?? Infinity);
+
+    const cronJob = new CronJob(data.params.expression, () => lock(this.wrapFunctionInTryCatchBlocks(data)));
+    const cronJobName = `${data.instance.constructor.name}::${data.methodName}`;
+
+    this.schedulerRegisty.addCronJob(cronJobName, cronJob);
+    cronJob.start();
+
+    this.logger.info(`CronJob ${cronJobName} started`);
```
davidleomay, line 60: I would remove this log 🤔

src/shared/services/dfx-cron.service.ts
```diff
@@ -0,0 +1,72 @@
...
+
+  private wrapFunctionInTryCatchBlocks(data: CronJobData) {
+    return async (...args: any) => {
+      if (data.params.process && DisabledProcess(data.params.process)) return;
+
+      if (data.params.useDelay ?? true) await Util.cronJobDelay(data.params.expression as CronExpression);
+
+      await data.methodRef.apply(data.instance, args);
+    };
+  }
```
davidleomay, line 71: This method name does not really make sense 😁

src/shared/services/dfx-cron.service.ts
bernd2022: I found this log useful as I saw at startup that the CronJobs had been started.

I have now removed the log output.

src/shared/services/dfx-cron.service.ts
bernd2022: Ok, great ... that was still an old name

PR #1956, opened 2025-02-09
[DEV-3121] add user guard

Yannick1712:
#### Pre-Release

- [ ] Check migrations
  - Login/logout
  - Buy/sell payment request
  - KYC page

#### Post-Release

- Test basic user operations

### DIFF
--- a/src/shared/auth/jwt-payload.interface.ts
+++ b/src/shared/auth/jwt-payload.interface.ts
@@ -1,4 +1,6 @@
 import { Blockchain } from 'src/integration/blockchain/shared/enums/blockchain.enum';
+import { UserDataStatus } from 'src/subdomains/generic/user/models/user-data/user-data.entity';
+import { UserStatus } from 'src/subdomains/generic/user/models/user/user.entity';
 import { UserRole } from './user-role.enum';
 
 export interface JwtPayload {
@@ -8,4 +10,6 @@ export interface JwtPayload {
   role: UserRole;
   blockchains?: Blockchain[];
   ip: string;
+  userStatus?: UserStatus;
+  userDataStatus?: UserDataStatus;
 }
--- a/src/shared/auth/jwt-payload.interface.ts
+++ b/src/shared/auth/jwt-payload.interface.ts
@@ -11,5 +11,5 @@ export interface JwtPayload {
   blockchains?: Blockchain[];
   ip: string;
   userStatus?: UserStatus;
-  userDataStatus?: UserDataStatus;
+  accountStatus?: UserDataStatus;
 }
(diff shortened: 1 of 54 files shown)


Yannick1712:
- [ ] add exception message in guard?
- [ ] add guard for all secured endpoints (User, Account, Admin)
- [ ] should a refund still be possible with blocked user?

src/shared/auth/user.guard.ts
```diff
@@ -0,0 +1,19 @@
+import { CanActivate, ExecutionContext, Injectable } from '@nestjs/common';
+import { UserDataStatus } from 'src/subdomains/generic/user/models/user-data/user-data.entity';
+import { UserStatus } from 'src/subdomains/generic/user/models/user/user.entity';
+
+@Injectable()
+export class UserGuard implements CanActivate {
+  async canActivate(context: ExecutionContext): Promise<boolean> {
+    const req = context.switchToHttp().getRequest();
```
davidleomay, line 8: Maybe we should rename this to `UserActiveGuard`?

src/shared/auth/user.guard.ts
```diff
@@ -0,0 +1,19 @@
...
+    const userStatus = req.user.userStatus;
+    const userDataStatus = req.user.userDataStatus;
+
+    return (
+      userStatus !== UserStatus.BLOCKED &&
+      userStatus !== UserStatus.DELETED &&
+      userDataStatus !== UserDataStatus.BLOCKED &&
+      userDataStatus !== UserDataStatus.DEACTIVATED
+    );
```
davidleomay, line 17: We should activate this in a second step. Because all existing JWTs will not have the new status fields.

src/shared/auth/jwt-payload.interface.ts
```diff
@@ -8,4 +10,6 @@ export interface JwtPayload {
   role: UserRole;
   blockchains?: Blockchain[];
   ip: string;
+  userStatus?: UserStatus;
+  userDataStatus?: UserDataStatus;
 }
```
davidleomay, line 15: `accountStatus`

src/shared/auth/user.guard.ts
Yannick1712: Why? That doesn't matter I would say 🤔 if userStatus or userDataStatus is undefined it's valid 👍 it's only invalid/forbidden if the status is blocked, so everything fine I would say, or?

davidleomay: True 👍  could do `[UserStatus.BLOCKED, UserStatus.DELETED].includes(userStatus) && ...`

src/shared/auth/user-active.guard.ts
```diff
@@ -0,0 +1,19 @@
...
+    const req = context.switchToHttp().getRequest();
+    const userStatus = req.user.userStatus;
+    const userDataStatus = req.user.userDataStatus;
+
+    return (
+      userStatus !== UserStatus.BLOCKED &&
+      userStatus !== UserStatus.DELETED &&
+      userDataStatus !== UserDataStatus.BLOCKED &&
+      userDataStatus !== UserDataStatus.DEACTIVATED
+    );
```
davidleomay, line 17: accountStatus?

davidleomay (approved):

PR #1972, opened 2025-02-18
[DEV-3626] bankTx senderChargeAmount

Yannick1712:
#### Pre-Release

- [ ] Check migrations
  - Login/logout
  - Buy/sell payment request
  - KYC page

#### Post-Release

- Test basic user operations

### DIFF
--- a/src/subdomains/supporting/bank-tx/bank-tx/services/sepa-parser.service.ts
+++ b/src/subdomains/supporting/bank-tx/bank-tx/services/sepa-parser.service.ts
@@ -76,6 +76,7 @@ export class SepaParser implements OnModuleInit {
         `CUSTOM/${file.BkToCstmrStmt.Stmt?.Acct?.Id?.IBAN}/${entry.BookgDt.Dt}/${entry.AddtlNtryInf}`;
 
       const creditDebitIndicator = this.toString(entry?.NtryDtls?.TxDtls?.CdtDbtInd);
+      const currency = this.toString(entry?.NtryDtls?.TxDtls?.Amt?.['@_Ccy']);
 
       let data: Partial<BankTx> = {};
       try {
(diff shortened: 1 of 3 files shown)


src/subdomains/supporting/bank-tx/bank-tx/services/sepa-parser.service.ts
```diff
@@ -120,31 +120,57 @@ export class SepaParser implements OnModuleInit {
...
+        chargeCurrency: Config.defaults.currency,
+        senderChargeAmount,
+        senderChargeAmountChf,
+        senderChargeCurrency: Config.defaults.currency,
+      };
 
     charges = (Array.isArray(charges) ? charges : [charges]).filter((c) => +c.Amt['#text'] !== 0);
 
+    const referenceAsset = await this.fiatService.getFiatByName(charges[0].Amt['@_Ccy']);
+
```
davidleomay, line 148: Don't we want the charge amount in the sender asset? It's not really defined, what the currency of the first charge is, is it?

src/subdomains/supporting/bank-tx/bank-tx/services/sepa-parser.service.ts
```diff
@@ -120,31 +120,57 @@ export class SepaParser implements OnModuleInit {
...
+      if (charge.CdtDbtInd === SepaCdi.DEBIT) {
+        chargeAmount += chargeReferencePrice.convert(amount, Config.defaultVolumeDecimal);
+        chargeAmountChf += chargeChfPrice.convert(amount, Config.defaultVolumeDecimal);
+        chargeCurrencies.add(currency);
+      } else {
+        senderChargeAmount += chargeReferencePrice.convert(amount, Config.defaultVolumeDecimal);
+        senderChargeAmountChf += chargeChfPrice.convert(amount, Config.defaultVolumeDecimal);
+        senderChargeCurrencies.add(currency);
+      }
```
davidleomay, line 164: I think it doesn't make sense to list all sender charge currencies if we anyway convert it to the reference currency?

src/subdomains/supporting/bank-tx/bank-tx/services/sepa-parser.service.ts
Yannick1712: Yeah good point, will change that 👍

davidleomay (approved):

PR #2066, opened 2025-04-07
[DEV-3672] fiatOutput update logic

Yannick1712:
#### Pre-Release

- [ ] Check migrations
  - Login/logout
  - Buy/sell payment request
  - KYC page

#### Post-Release

- Test basic user operations


Yannick1712:
- [ ] add migration

src/subdomains/supporting/fiat-output/fiat-output-job.service.ts
```diff
@@ -0,0 +1,261 @@
...
+      ) {
+        await this.fiatOutputRepo.update(
+          { id: In(currentBatch.map((f) => f.id)) },
+          { batchId: (await this.getLastBatchId()) + 1, batchAmount: currentBatchAmount * 100 },
+        );
+        currentBatch = [entity];
+      } else {
+        currentBatch.push(entity);
+      }
+    }
```
davidleomay, line 198: Is this supposed to only create batches if the total amount would be larger than `batchAmountLimit`?

davidleomay: Maybe it would be clever to have a `batches` array which we fill while looping through the entities and save them all in the end?

src/subdomains/supporting/fiat-output/fiat-output-job.service.ts
```diff
@@ -0,0 +1,261 @@
...
+
+        const bank = await this.bankService.getSenderBank(currency);
+
+        await this.fiatOutputRepo.update(entity.id, {
+          originEntityId: entity.type === FiatOutputType.BUY_FIAT ? entity.buyFiats[0].id : entity.buyCrypto.id,
+          accountIban: country.maerkiBaumannEnable ? bank?.iban : undefined,
+        });
+      } catch (e) {
+        this.logger.error(`Error in fiatOutput fillPreValutaDate job: ${entity.id}`, e);
+      }
```
davidleomay, line 103: Could also improve this log

src/subdomains/supporting/fiat-output/fiat-output-job.service.ts
```diff
@@ -0,0 +1,261 @@
...
+    private readonly logService: LogService,
+  ) {}
+
+  @DfxCron(CronExpression.EVERY_MINUTE, { process: Process.FIAT_OUTPUT, timeout: 1800 })
+  async fillFiatOutput() {
+    await this.fillPreValutaDate();
+    await this.fillIsReadyDate();
+    await this.setBatchId();
+    await this.setDates();
+    await this.bankTxSearch();
```
davidleomay, line 44: The job names are not really accurate 🤔  maybe they should be more about the included logic, e.g. more like:
- assignBankAccount
- setReady
- createBatches
- checkTransmission
- searchOutgoingBankTx

src/subdomains/supporting/fiat-output/fiat-output-job.service.ts
```diff
@@ -0,0 +1,261 @@
...
+            const ibanCountry = entity.iban.substring(0, 2);
+
+            if (
+              !entity.buyFiats.length ||
+              (entity.buyFiats?.[0]?.cryptoInput.isConfirmed &&
+                entity.buyFiats?.[0]?.cryptoInput.asset.blockchain &&
+                (liqBalance.asset.name !== 'CHF' || ['CH', 'LI'].includes(ibanCountry)))
+            )
+              await this.fiatOutputRepo.update(entity.id, { isReadyDate: new Date() });
+          }
```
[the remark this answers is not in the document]
Yannick1712: Yeah that's the logic of the current GS 👍 
I think it makes sense if you want to plan the payout and only waiting e.g. confirming cryptoInput

src/subdomains/supporting/log/log.service.ts
```diff
@@ -48,4 +48,25 @@ export class LogService {
 
     return this.logRepo.findOneBy({ id: Equal(maxId) });
   }
+
+  async getBankLogs(batchIds: string[]): Promise<Log[]> {
+    const query = this.logRepo
+      .createQueryBuilder('log')
+      .where('subsystem = UploadBank')
+      .andWhere('severity = :severity', { severity: LogSeverity.INFO });
```
davidleomay, line 56: I think we need ' around UploadBank?

src/subdomains/supporting/log/log.service.ts
```diff
@@ -48,4 +48,25 @@ export class LogService {
...
+      .where('subsystem = :subsystem', { subsystem: 'UploadBank' })
+      .andWhere('severity = :severity', { severity: LogSeverity.INFO });
+
+    query.andWhere(
+      new Brackets((query) =>
+        batchIds.forEach((id, index) => {
+          if (index === 0) {
+            query.where(`log.message LIKE '%${id}%'`);
+          } else {
+            query.orWhere(`log.message LIKE '%${id}%'`);
```
davidleomay, line 64: Hm, I think the best way to do this is probably
`query.where('log.message LIKE :message', { message: '%${id}%'});`

davidleomay (approved):

PR #2085, opened 2025-04-18
[DEV-3829] dfx safe step 1

Yannick1712:
#### Pre-Release

- [ ] Check migrations
  - Login/logout
  - Buy/sell payment request
  - KYC page

#### Post-Release

- Test basic user operations


Yannick1712:
- [ ] add migration
- [ ] add balance endpoints

src/subdomains/core/custody/services/custody.service.ts
```diff
@@ -0,0 +1,169 @@
...
+          .addSelect('SUM(custodyOrder.outputAmount)', 'withdrawal')
+          .where('custodyOrder.userId = :id', { id: order.user.id })
+          .andWhere('custodyOrder.status = :status', { status: CustodyOrderStatus.COMPLETED })
+          .andWhere('(custodyOrder.inputAssetId = :asset OR custodyOrder.outputAssetId = :asset)', {
+            asset: asset.id,
+          })
+          .getRawOne<{ deposit: number; withdrawal: number }>();
+
+        await this.updateCustodyBalance(custodyBalance, deposit - withdrawal + amount);
+      }
```
davidleomay, line 137: Isn't the amount already included in deposit - withdrawal?

davidleomay, line 137: Hm, this will update the balance for deposit transactions as soon as the user confirmed it 🤔  for withdraws we probably need to update the balance on order start, for deposits on order end.

src/subdomains/core/custody/entities/custody-balance.entity.ts
```diff
@@ -0,0 +1,20 @@
...
+import { Column, Entity, Index, ManyToOne } from 'typeorm';
+
+@Entity()
+@Index((custodyBalance: CustodyBalance) => [custodyBalance.user, custodyBalance.asset], { unique: true })
+export class CustodyBalance extends IEntity {
+  @Column({ type: 'float', default: 0 })
+  balance: number;
+
+  @Column({ type: 'float', default: 0 })
+  balanceInChf: number;
```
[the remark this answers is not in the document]
Yannick1712: Naming is a bit wrong I think, that would be the cummulated Purchase price

src/subdomains/core/custody/mappers/custody-asset-balance-dto.mapper.ts
```diff
@@ -0,0 +1,40 @@
...
+              ...group,
+              balance: Util.roundReadable(group.balance + dto.balance, AmountType.ASSET),
+              balanceInCurrency: Util.roundReadable(group.balanceInCurrency + dto.balanceInCurrency, AmountType.ASSET),
+            }
+          : dto,
+      );
+    }
+
+    return Array.from(groups.values());
+  }
```
davidleomay, line 39: Maybe this is simpler (remove the `mapCustodyBalance` method)?
```ts
const groups = Util.groupByAccessor(custodyBalances, (b) => b.asset.name);

return Array.from(groups.values())
  .map((g) => (
    const asset = g[0].asset;
    const price = Price.create(asset.name, currency.name, asset.getFiatPrice(currency));
    const balance = Util.sumObjValue(g, 'balance');

     const dto: CustodyAssetBalanceDto = {
      asset: { name: asset.name, description: asset.description },
      balance: Util.roundReadable(balance, AmountType.ASSET),
      balanceInCurrency: Util.roundReadable(price.convert(balance), AmountType.FIAT),
    };

    return Object.assign(new CustodyAssetBalanceDto(), dto);
  });
```

src/subdomains/core/custody/services/custody.service.ts
```diff
@@ -0,0 +1,171 @@
...
+
+    const custodyBalances = await this.custodyBalanceRepo.findBy({ user: { id: In(account.users.map((u) => u.id)) } });
+    const totalBalance = custodyBalances.reduce(
+      (prev, curr) =>
+        prev +
+        Price.create(curr.asset.name, account.currency.name, curr.asset.getFiatPrice(account.currency)).convert(
+          curr.balance,
+        ),
+      0,
+    );
```
davidleomay, line 151: Maybe it makes sense to map custody balances first and then sum up all values?

src/subdomains/core/custody/services/custody.service.ts
```diff
@@ -0,0 +1,171 @@
...
+  async startNextStep(step: CustodyOrderStep): Promise<void> {
+    const nextIndex = step.index + 1;
+    const order = step.order;
+    const nextStep = OrderConfig[order.type][nextIndex];
+
+    if (nextStep) {
+      await this.createStep(order, nextIndex, nextStep.command, nextStep.context);
+    } else {
+      const asset = order.inputAsset ?? order.outputAsset;
+      const amount = order.inputAmount ?? -order.outputAmount;
```
davidleomay, line 100: There might be orders with input and output asset/amount -> we need to update both balances?

src/subdomains/core/custody/mappers/custody-asset-balance-dto.mapper.ts
Yannick1712: Our "default" setting for our mappers is one method for one entity and a method for the batched entities, in case we need the method for one entity at some point, that's the reason why I added it in this way, but yeah we can do it one method too without `mapCustodyBalance` 👍

src/subdomains/core/custody/services/custody.service.ts
```diff
@@ -0,0 +1,160 @@
...
+              })
+              .orWhere('custodyOrder.type NOT IN (:...types)', {
+                types: [CustodyOrderType.DEPOSIT, CustodyOrderType.SAVING_DEPOSIT],
+              }),
+          ),
+        )
+        .andWhere('(custodyOrder.inputAssetId = :asset OR custodyOrder.outputAssetId = :asset)', {
+          asset: asset.id,
+        })
+        .getRawOne<{ deposit: number; withdrawal: number }>();
```
davidleomay, line 155: Separate queries

davidleomay (approved):

PR #2103, opened 2025-05-04
[DEV-3838] mod connection

Yannick1712:
#### Pre-Release

- [ ] Check migrations
  - Login/logout
  - Buy/sell payment request
  - KYC page

#### Post-Release

- Test basic user operations


src/subdomains/generic/user/models/user/user.service.ts
```diff
@@ -215,6 +217,11 @@ export class UserService {
     return user;
   }
 
+  async setModerator(user: User, moderator: Moderator): Promise<void> {
+    await this.userRepo.update(user.id, { usedRef: Config.moderators[moderator] });
+    await this.userDataService.updateUserDataInternal(user.userData, { moderator });
```
davidleomay, line 222: Only set moderator ref, if the user doesn't have one?

Yannick1712: Hehe yeah true 😁

davidleomay (approved):

PR #2127, opened 2025-05-12
[DEV-3847] add payment route endpoint

lapatric:
#### Pre-Release

- [ ] Check migrations
  - Login/logout
  - Buy/sell payment request
  - KYC page

#### Post-Release

- Test basic user operations

### DIFF
--- a/src/subdomains/core/payment-link/controllers/payment-link.controller.ts
+++ b/src/subdomains/core/payment-link/controllers/payment-link.controller.ts
@@ -13,7 +13,6 @@ import { JwtPayload } from 'src/shared/auth/jwt-payload.interface';
 import { RoleGuard } from 'src/shared/auth/role.guard';
 import { UserActiveGuard } from 'src/shared/auth/user-active.guard';
 import { UserRole } from 'src/shared/auth/user-role.enum';
-import { SellController } from 'src/subdomains/core/sell-crypto/route/sell.controller';
 import { SellService } from 'src/subdomains/core/sell-crypto/route/sell.service';
 import { UserDataService } from 'src/subdomains/generic/user/models/user-data/user-data.service';
 import { CreateInvoicePaymentDto } from '../dto/create-invoice-payment.dto';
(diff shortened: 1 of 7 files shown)


src/subdomains/core/route/route.controller.ts
```diff
@@ -37,6 +40,13 @@ export class RouteController {
     ]).then(([buy, sell, swap]) => ({ buy, sell, swap, crypto: swap }));
   }
 
+  @Get('payment/:id')
+  @ApiExcludeEndpoint()
+  async getPaymentRoute(@Param('id') id: string): Promise<SellDto> {
+    const sellRoute = await this.sellService.getPaymentRoute(id);
+    return this.sellController.toDto(sellRoute);
+  }
```
davidleomay, line 48: I think I would place this endpoint in `PaymentLinkController` under `GET paymentLink/recipient`.

Maybe also with query parameters instead of path parameter? I'm not sure 🤔

davidleomay: Maybe we can then somehow share the checks performed on the route? They are already in `PaymentLinkService`.

src/subdomains/core/payment-link/controllers/payment-link.controller.ts
```diff
@@ -125,6 +130,14 @@ export class PaymentLinkController {
 
   // --- PAYMENT --- //
 
+  @Get('recipient')
+  @ApiExcludeEndpoint()
+  @ApiQuery({ name: 'id', description: 'Route ID or label', required: true })
+  async getPaymentRecipient(@Query('id') id: string): Promise<SellDto> {
+    const sellRoute = await this.sellService.getPaymentRoute(id);
+    return this.sellController.toDto(sellRoute);
+  }
```
davidleomay, line 139: This sell DTO is a legacy thing that nobody uses 🤣  I would create a new one (e.g. `PaymentRecipient`) with the required information.

PR #2146, opened 2025-05-22
[DEV-3750] BtcService for blockchain registry

bernd2022:
#### Pre-Release

- [ ] Check migrations
  - Login/logout
  - Buy/sell payment request
  - KYC page

#### Post-Release

- Test basic user operations

### DIFF
--- a/src/integration/blockchain/ain/node/btc-client.ts
+++ b/src/integration/blockchain/ain/node/btc-client.ts
@@ -1,4 +1,8 @@
+import { Currency } from '@uniswap/sdk-core';
 import { Config } from 'src/config/config';
+import { Asset } from 'src/shared/models/asset/asset.entity';
+import { BlockchainTokenBalance } from '../../shared/dto/blockchain-token-balance.dto';
+import { BitcoinSignedTransactionResponse } from '../../shared/dto/signed-transaction-reponse.dto';
 import { NodeClient, NodeCommand } from './node-client';
 
 export interface TransactionHistory {
@@ -19,6 +23,10 @@ export interface TestMempoolResult {
   };
   'reject-reason': string;
 }
+
+type AddressInfoOuterArray = AddressInfoInnerArray[];
+type AddressInfoInnerArray = AddressInfoArray[];
+type AddressInfoArray = [string, number, string];
 
 export class BtcClient extends NodeClient {
   async send(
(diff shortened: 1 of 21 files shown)


src/integration/blockchain/ain/node/btc-client.ts
davidleomay, line 1: We could directly rename `ain` to `btc` (folder, module, ...)

src/integration/blockchain/shared/util/blockchain-client.ts
```diff
@@ -10,5 +10,5 @@ export abstract class BlockchainClient {
   abstract getTokenBalances(assets: Asset[], address?: string): Promise<BlockchainTokenBalance[]>;
   abstract isTxComplete(txHash: string, confirmations?: number): Promise<boolean>;
   abstract getToken(asset: Asset): Promise<Currency>;
-  abstract sendSignedTransaction(tx: string): Promise<SignedTransactionResponse>;
+  abstract sendSignedTransaction(tx: string): Promise<EvmSignedTransactionResponse>;
```
davidleomay, line 13: BtcClient also extends BlockchainClient but does not return EvmSignedTransactionResponse?

src/integration/blockchain/shared/services/blockchain-registry.service.ts
```diff
@@ -22,9 +24,10 @@ export class BlockchainRegistryService {
     private readonly polygonService: PolygonService,
     private readonly baseService: BaseService,
     private readonly moneroService: MoneroService,
+    private readonly btcService: BtcService,
   ) {}
 
-  getClient(blockchain: Blockchain): EvmClient | MoneroClient {
+  getClient(blockchain: Blockchain): EvmClient | MoneroClient | BtcClient {
     return this.getService(blockchain).getDefaultClient();
   }
```
davidleomay, line 32: This will throw an error for bitcoin, as the type is not defined. Maybe we should by default return the Input client?

src/integration/blockchain/ain/node/btc.service.ts
davidleomay, line 1: I think all the subscription logic can be removed. This was used when we had active and passive nodes to improve availability.
Also `addNodeClientPair` does not make sense anymore, it's just a single node.

src/subdomains/core/payment-link/services/payment-quote.service.ts
```diff
@@ -456,8 +456,11 @@ export class PaymentQuoteService {
         return;
       }
 
-      const txId = await this.payoutBitcoinService.sendRawTransaction(transferInfo.hex);
-      txId ? quote.txInMempool(txId) : quote.txFailed('Transaction failed');
+      const transactionResponse = await this.payoutBitcoinService.sendSignedTransaction(transferInfo.hex);
+
+      transactionResponse.error
+        ? quote.txFailed(transactionResponse.error.message)
+        : quote.txInMempool(transactionResponse.hash);
```
davidleomay, line 463: Why not do this directly on the client (as for EVM)?

PR #2263, opened 2025-07-06
[DEV-4103] bankData status

Yannick1712:
#### Pre-Release

- [ ] Check migrations
  - Login/logout
  - Buy/sell payment request
  - KYC page

#### Post-Release

- Test basic user operations

### DIFF
--- a/src/subdomains/generic/kyc/enums/review-status.enum.ts
+++ b/src/subdomains/generic/kyc/enums/review-status.enum.ts
@@ -1,0 +1,17 @@
+export enum ReviewStatus {
+  NOT_STARTED = 'NotStarted',
+  IN_PROGRESS = 'InProgress',
+  FINISHED = 'Finished',
+  EXTERNAL_REVIEW = 'ExternalReview',
+  INTERNAL_REVIEW = 'InternalReview',
+  MANUAL_REVIEW = 'ManualReview',
+  FAILED = 'Failed',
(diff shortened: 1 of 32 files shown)


Yannick1712:
- [ ] add migration

src/subdomains/generic/kyc/enums/review-status.enum.ts
```diff
@@ -0,0 +1,17 @@
...
+  FAILED = 'Failed',
+  COMPLETED = 'Completed',
+  PARTIALLY_APPROVED = 'PartiallyApproved',
+  CANCELED = 'Canceled',
+  IGNORED = 'Ignored',
+  OUTDATED = 'Outdated',
+  DATA_REQUESTED = 'DataRequested',
+  PAUSED = 'Paused',
+  ON_HOLD = 'OnHold',
+}
```
davidleomay, line 17: I'm not sure if it makes sense to combine the two enums into one. There are many status that are only used for KYC so it might be quite confusing for bank datas 🤔

davidleomay: Haha, yes. As an outsider it's easier to see what's going on when using separate enums. But we can leave it as its is, if you prefer 😉

davidleomay (approved):

davidleomay (approved):

PR #2266, opened 2025-07-07
[DEV-4130] Expiration timers

bernd2022:
#### Pre-Release

- [ ] Check migrations
  - Login/logout
  - Buy/sell payment request
  - KYC page

#### Post-Release

- Test basic user operations

### DIFF
--- a/src/subdomains/core/payment-link/services/payment-link-payment.service.ts
+++ b/src/subdomains/core/payment-link/services/payment-link-payment.service.ts
@@ -215,10 +215,9 @@ export class PaymentLinkPaymentService {
       }, paymentLink.configObj.autoConfirmSecs * 1000);
     }
 
-    void Util.timeout(
-      this.expirePaymentIfPending(payment.uniqueId),
-      payment.expiryDate.getTime() - new Date().getTime(),
-    );
+    setTimeout(async () => {
+      await this.expirePaymentIfPending(payment.uniqueId);
+    }, payment.expiryDate.getTime() - new Date().getTime());
 
     return savedPayment;
   }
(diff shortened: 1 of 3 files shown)


src/subdomains/core/payment-link/services/payment-link.service.ts
```diff
@@ -286,6 +286,11 @@ export class PaymentLinkService {
       transferAmounts: actualQuote.transferAmountsForPayRequest,
     };
 
+    await Util.timeout(
+      this.paymentLinkPaymentService.processExpiredPayments(),
+      Util.secondsDiff(actualQuote.expiryDate) * 1000,
+    );
```
davidleomay, line 292: I think it doesn't make sense to wait for this timeout when creating a payment request.

davidleomay: We should start the timer when creating a payment, not for every payment request.

davidleomay: `Util.secondsDiff(actualQuote.expiryDate)` will give a negative number?

src/subdomains/core/payment-link/services/payment-link-payment.service.ts
```diff
@@ -55,6 +55,12 @@ export class PaymentLinkPaymentService {
 
   // --- JOBS --- //
   async processExpiredPayments(): Promise<void> {
+    await this.doProcessExpiredPayments();
+    await this.paymentActivationService.processExpiredActivations();
+    await this.paymentQuoteService.processExpiredQuotes();
+  }
```
davidleomay, line 61: The code below will already cancel all quotes/activations for the expired payment. No need to process them separately. Keep this only for the cron job.
```ts
      await this.doSave(payment.expire(), true);
      await this.cancelQuotesForPayment(payment);
```
-> Keep the cron job as it is (with prolonged execution period)
-> Just call the code above when a payment expires (maybe move to a method)

davidleomay (approved):

PR #2280, opened 2025-07-11
[NOTASK] Improve bankData update, [DEV-3990]: Payment Link modes

TaprootFreak:
#### Pre-Release

- [ ] Check migrations
  - Login/logout
  - Buy/sell payment request
  - KYC page

#### Post-Release

- Test basic user operations

### DIFF
--- a/src/subdomains/core/payment-link/services/payment-link-payment.service.ts
+++ b/src/subdomains/core/payment-link/services/payment-link-payment.service.ts
@@ -24,6 +24,7 @@ import { PaymentDevice, PaymentLinkPayment } from '../entities/payment-link-paym
 import { PaymentLink } from '../entities/payment-link.entity';
 import { PaymentQuote } from '../entities/payment-quote.entity';
 import {
+  PaymentLinkMode,
   PaymentLinkPaymentMode,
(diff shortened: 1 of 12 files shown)


src/subdomains/core/payment-link/services/payment-link-payment.service.ts
```diff
@@ -184,11 +185,16 @@ export class PaymentLinkPaymentService {
...
+      if (allPayments.length > 0) throw new ConflictException('Single payment link can only have one payment');
+    }
+
     if (dto.externalId) {
-      const exists = await this.paymentLinkPaymentRepo.existsBy({
-        externalId: dto.externalId,
-        link: { id: paymentLink.id },
-      });
+      const exists = allPayments.some((p) => p.externalId === dto.externalId);
       if (exists) throw new ConflictException('Payment already exists');
```
davidleomay, line 198: I think I would split this in two DB requests. And only if payment link mode is SINGLE, check if there are already payments with an `exists` call. Loading all payments for a link might result in a huge amount of data.

davidleomay (approved):

PR #2491, opened 2025-10-07
[DEV-3957] partner fee

Yannick1712:
#### Pre-Release

- [ ] Check migrations
  - Login/logout
  - Buy/sell payment request
  - KYC page

#### Post-Release

- Test basic user operations

### DIFF
--- a/src/subdomains/supporting/payment/dto/transaction-helper/quote-error.util.ts
+++ b/src/subdomains/supporting/payment/dto/transaction-helper/quote-error.util.ts
@@ -65,7 +65,18 @@ export class QuoteErrorUtil {
 
   static createErrorQuote(error: QuoteError | QuoteException): ErrorQuote {
     const quoteError = error instanceof QuoteException ? error.error : error;
-    const emptyFee: FeeDto = { rate: 0, fixed: 0, network: 0, min: 0, dfx: 0, bank: 0, bankFixed: 0, bankVariable: 0, total: 0, platform: 0 };
+    const emptyFee: FeeDto = {
+      rate: 0,
+      fixed: 0,
+      network: 0,
+      min: 0,
+      dfx: 0,
+      bank: 0,
+      bankFixed: 0,
+      bankVariable: 0,
+      total: 0,
+      platform: 0,
+    };
 
     return {
       feeAmount: 0,


davidleomay (commented):
- [x] Improve naming

davidleomay (changes requested):
- [ ] Implement partner fee payout

src/subdomains/core/buy-crypto/process/services/buy-crypto.service.ts
```diff
@@ -896,6 +899,18 @@ export class BuyCryptoService {
     return { volume: volume ?? 0, credit: credit ?? 0 };
   }
 
+  async getPartnerFeeRefVolume(ref: string): Promise<{ volume: number; credit: number }> {
+    const { volume, credit } = await this.buyCryptoRepo
+      .createQueryBuilder('buyCrypto')
+      .select('SUM((partnerFeeAmount * (amountInEur/amountInChf )) / (refProvision * 0.01))', 'volume')
+      .addSelect('SUM(partnerFeeAmount * (amountInEur/amountInChf ))', 'credit')
```
davidleomay: Isn't the volume just `amountInEur`?

Yannick1712: Nope the volume is always calculated with the refProvision of the user, normally we would take the amountInEur and calculate the credit with it but in this scenario the credit is given with the partnerFeeAmount and we have to calculate the right volume for it to match the ref calculation

src/subdomains/generic/user/models/user/user.entity.ts
```diff
@@ -226,6 +225,14 @@ export class User extends IEntity {
...
     return this.status === UserStatus.DELETED;
   }
+
+  get completeRefVolume(): number {
+    return this.refVolume + this.partnerRefVolume;
+  }
+
+  get completeRefCredit(): number {
+    return this.refCredit + this.partnerRefCredit;
+  }
```
davidleomay: I would rather use `total` instead of `complete` (e.g. `totalRefVolume`)

src/subdomains/core/sell-crypto/process/buy-fiat.entity.ts
```diff
@@ -87,6 +88,9 @@ export class BuyFiat extends IEntity {
   @Column({ length: 256, nullable: true })
   usedRef?: string;
 
+  @Column({ length: 256, nullable: true })
+  usedPartnerFeeRef?: string;
```
davidleomay: Same here?

PR #2533, opened 2025-10-30
[DEV-4461] fix compliance search middle name bug

Yannick1712:
#### Pre-Release

- [ ] Check migrations
  - Login/logout
  - Buy/sell payment request
  - KYC page

#### Post-Release

- Test basic user operations

### DIFF
--- a/src/subdomains/generic/user/models/user-data/user-data.service.ts
+++ b/src/subdomains/generic/user/models/user-data/user-data.service.ts
@@ -44,7 +44,7 @@ import { MailContext } from 'src/subdomains/supporting/notification/enums';
 import { SpecialExternalAccountService } from 'src/subdomains/supporting/payment/services/special-external-account.service';
 import { TransactionService } from 'src/subdomains/supporting/payment/services/transaction.service';
 import { transliterate } from 'transliteration';
-import { Equal, FindOptionsRelations, FindOptionsWhere, ILike, In, IsNull, Not } from 'typeorm';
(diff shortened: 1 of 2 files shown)


src/subdomains/generic/user/models/user-data/user-data.service.ts
```diff
@@ -180,12 +180,28 @@ export class UserDataService {
...
+
     return this.userDataRepo.find({
       where: [
         { ...where, firstname: ILike(search) },
         { ...where, surname: ILike(search) },
         { ...where, verifiedName: ILike(search) },
         { ...where, organization: { name: ILike(search) } },
+        ...nameSplitRequest,
       ],
     });
```
davidleomay: I think I would do it like this:
```TS
// util method
Contains(search: string): FindOperator<string> {
   return ILike(`%${search}%`);
}

// code
const where = { status: Not(UserDataStatus.MERGED) };
const wheres = [
        { ...where, firstname: Contains(name) },
        { ...where, surname: Contains(name) },
        { ...where, verifiedName: Contains(name) },
        { ...where, organization: { name: Contains(name) } },
];

const nameParts = name.split(' ');
const first = name.shift();
const last = name.pop();

if (last) wheres.push({ ...where, firstname: Contains(first), surname: Contains([...nameParts, last].join(' '))});
if (nameParts.length) wheres.push({ ...where, firstname: Contains([first, ...nameParts].join(' ')), surname: Contains(last)})

return this.userDataRepo.find({ where: wheres });
```
WDYT?

PR #2550, opened 2025-11-11
[DEV-4473] OCP crypto payout

davidleomay:
#### Pre-Release

- [ ] Check migrations
  - Login/logout
  - Buy/sell payment request
  - KYC page

#### Post-Release

- Test basic user operations

### DIFF
--- a/src/subdomains/supporting/address-pool/route/deposit-route.service.ts
+++ b/src/subdomains/supporting/address-pool/route/deposit-route.service.ts
@@ -1,7 +1,7 @@
-import { BadRequestException, Injectable, NotFoundException } from '@nestjs/common';
+import { Injectable, NotFoundException } from '@nestjs/common';
 import { merge } from 'lodash';
 import { Blockchain } from 'src/integration/blockchain/shared/enums/blockchain.enum';
-import { DfxLogger } from 'src/shared/services/dfx-logger';
+import { Util } from 'src/shared/utils/util';
 import { PaymentLink } from 'src/subdomains/core/payment-link/entities/payment-link.entity';
 import { FindOneOptions, In } from 'typeorm';
 import { DepositRoute } from './deposit-route.entity';
@@ -9,17 +9,15 @@ import { DepositRouteRepository } from './deposit-route.repository';
 
 @Injectable()
 export class DepositRouteService {
-  private readonly logger = new DfxLogger(DepositRouteService);
-
   constructor(private readonly depositRouteRepo: DepositRouteRepository) {}
 
   async get(userId: number, id: number): Promise<DepositRoute> {
-    const routes = await this.depositRouteRepo.findOne({
+    const route = await this.depositRouteRepo.findOne({
       where: { id, user: { id: userId } },
       relations: { user: { userData: true } },
     });
-    if (!routes) throw new NotFoundException('Route not found');
-    return routes;
+    if (!route) throw new NotFoundException('Route not found');
+    return route;
   }
 
   async getById(id: number, options?: FindOneOptions<DepositRoute>): Promise<DepositRoute> {
@@ -27,7 +25,7 @@ export class DepositRouteService {
     return this.depositRouteRepo.findOne(merge(defaultOptions, options));
   }
 
-  async getLatest(userId: number): Promise<DepositRoute | null> {
+  async getLatest(userId: number): Promise<DepositRoute | undefined> {
     return this.depositRouteRepo.findOne({
       where: { user: { id: userId } },
       relations: { user: { userData: true } },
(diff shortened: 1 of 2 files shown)


src/subdomains/core/buy-crypto/process/services/buy-crypto-preparation.service.ts
```diff
@@ -302,6 +305,119 @@ export class BuyCryptoPreparationService {
...
+        isComplete: false,
+        inputReferenceAmount: Not(IsNull()),
+        cryptoInput: { paymentLinkPayment: { id: Not(IsNull()) }, status: PayInStatus.COMPLETED },
+      },
+
+      relations: {
+        bankTx: true,
+        checkoutTx: true,
+        cryptoInput: true,
+        buy: true,
```
Yannick1712: Same

src/subdomains/supporting/address-pool/route/deposit-route.service.ts
```diff
@@ -0,0 +1,111 @@
...
+  async getLatest(userId: number): Promise<DepositRoute | null> {
+    return this.depositRouteRepo.findOne({
+      where: { user: { id: userId } },
+      relations: { user: { userData: true } },
+      order: { created: 'DESC' },
+    });
+  }
+
+  async getByLabel(userId: number, label: string, options?: FindOneOptions<DepositRoute>): Promise<DepositRoute> {
+    const defaultOptions = {
```
Yannick1712: Why do we need `defaultOptions` as variable?

src/subdomains/supporting/address-pool/route/deposit-route.service.ts
```diff
@@ -0,0 +1,111 @@
...
+    routeIdOrLabel: string,
+    externalIds?: string[],
+    ids?: number[],
+  ): Promise<PaymentLink[]> {
+    const route = await this.getPaymentRoute(routeIdOrLabel, {
+      relations: { paymentLinks: true },
+      where: {
+        paymentLinks: [
+          ...(externalIds?.length ? [{ externalId: In(externalIds) }] : []),
+          ...(ids?.length ? [{ id: In(ids) }] : []),
```
Yannick1712: We should consider if the `In`-Method is the right, its limit is not so high, we will get errors if `externalIds` or `ids` are to long 🤔

src/subdomains/supporting/address-pool/route/deposit-route.service.ts
```diff
@@ -0,0 +1,111 @@
...
+  }
+
+  async getPaymentRoute(idOrLabel: string, options?: FindOneOptions<DepositRoute>): Promise<DepositRoute> {
+    const isRouteId = !isNaN(+idOrLabel);
+    const route = isRouteId
+      ? await this.getById(+idOrLabel, options)
+      : await this.getByLabel(undefined, idOrLabel, options);
+
+    try {
+      this.validateLightningRoute(route);
```
Yannick1712: That's a bit strange, `validateLightningRoute` method has only 2 if, that makes no sense to use the method here imo 🤔 
I would make a if statement here and directly throw the right exception with the right text

src/subdomains/supporting/address-pool/route/deposit-route.service.ts
```diff
@@ -0,0 +1,111 @@
...
+    } catch (e) {
+      this.logger.verbose(`Failed to validate route ${idOrLabel}:`, e);
```
Yannick1712: Do we need this log? 🤔

src/subdomains/supporting/address-pool/route/deposit-route.service.ts
```diff
@@ -0,0 +1,111 @@
...
+      where: {
+        paymentLinks: [
+          ...(externalIds?.length ? [{ externalId: In(externalIds) }] : []),
+          ...(ids?.length ? [{ id: In(ids) }] : []),
+        ],
+      },
+      order: { paymentLinks: { created: 'ASC' } },
+    });
+
+    return Array.from(new Map((route.paymentLinks || []).map((l) => [l.id, l])).values());
```
Yannick1712: Could we use `Util.toUniqueList`?

src/subdomains/supporting/address-pool/route/deposit-route.service.ts
```diff
@@ -0,0 +1,111 @@
...
+      relations: { user: { userData: true } },
+    };
+    return this.depositRouteRepo.findOne(merge(defaultOptions, options));
+  }
+
+  validateLightningRoute(route: DepositRoute): void {
+    if (!route) throw new NotFoundException('Route not found');
+    if (route.deposit.blockchains !== Blockchain.LIGHTNING)
+      throw new BadRequestException('Only Lightning routes are allowed');
+  }
```
Yannick1712: Do we really need that method? It is only used on 2 other locations? For me its a bit strange with only 2 if statements in it 🤔

src/subdomains/supporting/address-pool/route/deposit-route.service.ts
```diff
@@ -0,0 +1,111 @@
...
+  async get(userId: number, id: number): Promise<DepositRoute> {
+    const routes = await this.depositRouteRepo.findOne({
+      where: { id, user: { id: userId } },
+      relations: { user: { userData: true } },
+    });
+    if (!routes) throw new NotFoundException('Route not found');
+    return routes;
+  }
+
+  async getById(id: number, options?: FindOneOptions<DepositRoute>): Promise<DepositRoute> {
```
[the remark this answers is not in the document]
davidleomay: The call structure is a bit nested, would require code duplications 🤔

PR #2561, opened 2025-11-18
[NOTASK]&[DEV-4489]&[DEV-4492]&[DEV-4447]&[DEV-4488] Improvements, fixes, ipLog amlCheck

Yannick1712:
#### Pre-Release

- [ ] Check migrations
  - Login/logout
  - Buy/sell payment request
  - KYC page

#### Post-Release

- Test basic user operations


src/subdomains/supporting/bank/bank-account/is-dfx-iban.validator.ts
```diff
@@ -72,6 +72,7 @@ export class IsDfxIbanValidator implements ValidatorConstraintInterface {
 
   defaultMessage(args: ValidationArguments): string | undefined {
     const iban = args.value;
+    if (!iban) return 'IBAN required';
```
davidleomay: @Yannick1712 I would just return undefined, if no value

PR #2620, opened 2025-12-11
Set bankFixed fee to 1 for Bank type fees

TaprootFreak:
## Summary
Migration to set `fee.fixed = 1` for all Bank type fee entries.

## Related
Companion PR to #2619 (Remove currency conversion for bankFixed fee)

## Migration
Sets `fixed = 1` for all fees where `type = 'Bank'`


davidleomay:
Done manually

PR #2650, opened 2025-12-16
[DEV-4518] bankRelation fiatOutput

Yannick1712:
#### Pre-Release

- [ ] Check migrations
  - Login/logout
  - Buy/sell payment request
  - KYC page

#### Post-Release

- Test basic user operations


src/subdomains/supporting/fiat-output/fiat-output-job.service.ts
davidleomay: Remove all methods that are no longer in use, e.g.
- getAllYapealIbans
- getIbansByName
- getAllActiveVirtualIbans

PR #2659, opened 2025-12-17
Add AML error for intermediary banks without sender name

TaprootFreak:
## Summary
- Add new `INTERMEDIARY_WITHOUT_SENDER` AML error for transactions where the bank name is an intermediary (Wise/Revolut) but no actual sender name (`ultimateName`) is provided
- Automatically triggers `CheckStatus.FAIL` (refund) when detected
- Multi-account bank names are loaded dynamically from `special_external_account` table (type `MultiAccountBankName`) with caching

## Changes
- `aml-reason.enum.ts` / `aml-error.enum.ts`: New error and reason
- `aml-helper.service.ts`: Check logic in bank transaction flow
- `special-external-account.service.ts`: New `getMultiAccountBankNames()` method
- `aml.service.ts`: Load and pass bank names through AML check
- `buy-crypto.entity.ts` / `buy-crypto-preparation.service.ts`: Pass parameter through chain
- `sift.dto.ts` / `transaction.dto.ts`: Required mappings
- i18n translations (DE, EN, FR, IT, ES, PT)

## Test Plan
- [ ] Verify error triggers when `bank_tx.name` = "Wise" or "Revolut Bank UAB" AND `bank_tx.ultimateName` = NULL
- [ ] Verify error does NOT trigger when `ultimateName` has a value
- [ ] Verify error does NOT trigger for other bank names
- [ ] Verify chargeback email contains correct reason text

### DIFF
--- a/src/subdomains/core/aml/enums/aml-error.enum.ts
+++ b/src/subdomains/core/aml/enums/aml-error.enum.ts
@@ -62,7 +62,7 @@ export enum AmlError {
   BANK_RELEASE_DATE_MISSING = 'BankReleaseDateMissing',
   IP_COUNTRY_MISMATCH = 'IpCountryMismatch',
   TRADE_APPROVAL_DATE_MISSING = 'TradeApprovalDateMissing',
-  INTERMEDIARY_WITHOUT_SENDER = 'IntermediaryWithoutSender',
+  BANK_TX_CUSTOMER_NAME_MISSING = 'BankTxCustomerNameMissing',
 }
 
 export const DelayResultError = [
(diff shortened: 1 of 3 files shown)


src/subdomains/core/aml/enums/aml-error.enum.ts
```diff
@@ -62,6 +62,7 @@ export enum AmlError {
   BANK_RELEASE_DATE_MISSING = 'BankReleaseDateMissing',
   IP_COUNTRY_MISMATCH = 'IpCountryMismatch',
   TRADE_APPROVAL_DATE_MISSING = 'TradeApprovalDateMissing',
+  INTERMEDIARY_WITHOUT_SENDER = 'IntermediaryWithoutSender',
```
Yannick1712: The error should be renamed to `CustomerNameMissing` or something like that. This would be better for the amlError, the amlReason is perfectly fine with `IntermediaryWithoutSender` and should not be changed

PR #2681, opened 2025-12-22
Add bank account details to refund endpoint

TaprootFreak:
## Summary
- Extend `RefundDataDto` with bank account fields: name, address, houseNumber, zip, city, country, iban, bic
- Extract bank details from `BankTx` entity in `getRefundData` method
- Add helper method `getBankTxFromRefundEntity` to handle different refund entity types

## Field Mapping
| BankTx Field | RefundDataDto Field |
|--------------|---------------------|
| name | name |
| addressLine1 | address |
| addressLine2 | city |
| country | country |
| iban | iban |
| bic | bic |

## Related PRs
- packages: feature/refund-bank-details
- services: feature/refund-bank-details

### DIFF
--- a/src/subdomains/core/history/dto/refund-data.dto.ts
+++ b/src/subdomains/core/history/dto/refund-data.dto.ts
@@ -14,28 +14,7 @@ export class RefundFeeDto {
   dfx: number;
 }
 
-export class RefundDataDto {
-  @ApiProperty({ description: 'Expiry date of the refund data' })
-  expiryDate: Date;
-
-  @ApiProperty({ type: RefundFeeDto, description: 'Refund fees' })
-  fee: RefundFeeDto;
-
-  @ApiProperty()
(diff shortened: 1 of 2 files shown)


src/subdomains/core/history/dto/refund-data.dto.ts
```diff
@@ -34,5 +34,29 @@ export class RefundDataDto {
...
+  city?: string;
+
+  @ApiPropertyOptional({ description: 'Country code' })
+  country?: string;
+
+  @ApiPropertyOptional({ description: 'IBAN' })
+  iban?: string;
+
+  @ApiPropertyOptional({ description: 'BIC/SWIFT code' })
+  bic?: string;
```
Yannick1712: Maybe would be better to store the data in its own dto, something like `bankData` or something like that

PR #2701, opened 2025-12-24
[NO-TASK] Add monthly volume

Kolibri1990:
#### Pre-Release

- [ ] Check migrations
  - Login/logout
  - Buy/sell payment request
  - KYC page

#### Post-Release

- Test basic user operations


src/subdomains/generic/user/models/user-data/user-data.entity.ts
davidleomay: Add migration

PR #2879, opened 2026-01-08
feat(aml): add NAME_TOO_SHORT check for bank payout validation

TaprootFreak:
## Summary
- Add AML check to reject transactions where user's name has fewer than 4 letters
- Banks (e.g. Yapeal) require minimum name length for processing payouts
- Example: "M K" (2 letters) → `amlCheck: FAIL`, `amlReason: NameTooShort`

## Changes
- Add `AmlError.NAME_TOO_SHORT` with `CRUCIAL` type and `FAIL` status
- Add `AmlReason.NAME_TOO_SHORT`
- Add `countLetters()` helper counting only alphabetic characters (incl. ä, ö, ü, ß, etc.)
- Add validation in `getAmlErrors()` after existing `NAME_MISSING` check
- Update `SiftAmlDeclineMap` and `TransactionReasonMapper`

## Test plan
- [ ] Test with user name "M K" (2 letters) → should get `amlCheck: FAIL`
- [ ] Test with user name "Li Wu" (4 letters) → should pass this check
- [ ] Test with user name "Hans Müller" (10 letters) → should pass
- [ ] Verify transaction shows `NameTooShort` reason in admin panel

### DIFF
--- a/src/subdomains/core/aml/services/aml-helper.service.ts
+++ b/src/subdomains/core/aml/services/aml-helper.service.ts
@@ -615,6 +615,6 @@ export class AmlHelperService {
   }
 
   private static countLetters(str: string): number {
-    return str.replace(/[^a-zA-ZäöüÄÖÜàáâãéèêëìíîïòóôõùúûüñçßÀÁÂÃÈÉÊËÌÍÎÏÒÓÔÕÙÚÛÜÑÇ]/g, '').length;
+    return str.replace(/[^\p{L}]/gu, '').length;
   }
 }


src/subdomains/core/aml/enums/aml-reason.enum.ts
```diff
@@ -39,6 +39,7 @@ export enum AmlReason {
   BANK_RELEASE_PENDING = 'BankReleasePending',
   VIRTUAL_IBAN_USER_MISMATCH = 'VirtualIbanUserMismatch',
   INTERMEDIARY_WITHOUT_SENDER = 'IntermediaryWithoutSender',
+  NAME_TOO_SHORT = 'NameTooShort',
 }
```
davidleomay: @Yannick1712 do we need to add mail texts when adding a new AmlReason?

PR #2919, opened 2026-01-13
fix: check unlocked balance per group for Monero payouts

TaprootFreak:
## Summary
- Check unlocked balance **per 15-order group** instead of for all orders at once
- Ensures at least the first group gets processed when sufficient funds exist
- Remaining groups retry automatically in subsequent cronjob runs (every 30s)

## Problem
Previously, if 45 orders required 90 XMR but only 35 XMR was unlocked, **all 45 orders were skipped** - even though the first group (30 XMR) could have been processed.

## Solution
Move balance check inside the group loop with fresh balance query per group. This guarantees maximum throughput given the 10-block lock constraint on Monero change outputs.

## Test plan
- [ ] Deploy to DEV and verify groups are processed sequentially
- [ ] Check logs show "Insufficient unlocked balance for XMR group" with Need/Have amounts
- [ ] Verify orders in later groups retry in subsequent cronjob runs

### DIFF
--- a/src/subdomains/supporting/payout/strategies/payout/impl/monero.strategy.ts
+++ b/src/subdomains/supporting/payout/strategies/payout/impl/monero.strategy.ts
@@ -42,58 +42,45 @@ export class MoneroStrategy extends BitcoinBasedStrategy {
   }
 
   protected async doPayoutForContext(context: PayoutOrderContext, orders: PayoutOrder[]): Promise<void> {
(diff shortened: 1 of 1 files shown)


src/subdomains/supporting/payout/strategies/payout/impl/monero.strategy.ts
davidleomay: Improve group creation

bernd2022 (approved):

PR #2934, opened 2026-01-13
feat: add multi-app Application Insights log access

TaprootFreak:
## Summary
- Add support for querying logs from all container apps via the debug endpoint
- Currently only DFX API logs are accessible; this extends access to JuiceDollar, dEuro, Ponder apps etc.
- Useful for investigating high log volumes and reducing Azure costs

## Changes
- Add `ContainerApp` enum with all container apps
- Extend config with app-specific Application Insights IDs
- Add optional `app` parameter to `LogQueryDto`
- Update `AppInsightsQueryService` to use dynamic app ID
- Extend `log-debug.sh` script with `-a/--app` parameter

## Usage
```bash
# DFX API (default)
./scripts/log-debug.sh exceptions

# JuiceDollar API
./scripts/log-debug.sh exceptions -a juicedollarApi

# dEuro API
./scripts/log-debug.sh traces "error" -a deuroApi -h 24

# dEuro Monitoring
./scripts/log-debug.sh exceptions -a deuroMonitoring
```

## Configuration Required
Add the Application Insights App IDs to the environment:
```env
APPINSIGHTS_JUICEDOLLAR_API_APP_ID=<App ID from Azure Portal>
APPINSIGHTS_DEURO_API_APP_ID=<App ID from Azure Portal>
APPINSIGHTS_DEURO_MONITORING_APP_ID=<App ID from Azure Portal>
APPINSIGHTS_JUICESWAP_PONDER_APP_ID=<App ID from Azure Portal>
APPINSIGHTS_DEURO_PONDER_APP_ID=<App ID from Azure Portal>
APPINSIGHTS_REALUNIT_PONDER_APP_ID=<App ID from Azure Portal>
```

## Test plan
- [ ] Deploy to DEV and configure App IDs
- [ ] Test `./scripts/log-debug.sh exceptions -a juicedollarApi`
- [ ] Verify logs from different container apps are accessible

### DIFF
--- a/src/integration/infrastructure/app-insights-query.service.ts
+++ b/src/integration/infrastructure/app-insights-query.service.ts
@@ -18,10 +18,10 @@ export class AppInsightsQueryService {
   constructor(private readonly http: HttpService) {}
 
   async query(kql: string, timespan?: string, app?: ContainerApp): Promise<AppInsightsQueryResponse> {
-    const { apiKey, apps } = Config.azure.appInsights;
+    const { appId: defaultAppId, apiKey, apps } = Config.azure.appInsights;
 
     // Use specified app or default to dfxApi
-    const appId = app ? apps[app] : apps.dfxApi ?? Config.azure.appInsights.appId;
+    const appId = app ? apps[app] : defaultAppId;
 
     if (!appId || !apiKey) {
       throw new Error(app ? `App insights config missing for ${app}` : 'App insights config missing');


davidleomay:
The other container apps (JuiceDollar, dEuro, Ponder) don't use Application Insights - they just log to the Log Analytics
  workspace. This means they don't have App IDs, so the current approach won't work.

  To query logs from these services, we'd need to:
  1. Query the Log Analytics API instead of the Application Insights API
  2. Filter by container app name in the KQL query, e.g.:
  ContainerAppConsoleLogs
  | where ContainerAppName == "juicedollar-api"
  | where Log contains "error"

  This would require a different service/endpoint that uses the Log Analytics workspace ID and API key, not the Application Insights
   credentials.

PR #2962, opened 2026-01-16
feat(payout): add auto-retry for stuck payout orders

TaprootFreak:
## Summary
- Adds automatic retry for payout orders stuck in `PAYOUT_PENDING` status longer than 10 minutes
- Handles expired Flashbots transactions that were never mined
- Configurable timeout via `PAYOUT_STUCK_TIMEOUT_MINUTES` environment variable

## Problem
When transactions are sent via Flashbots Protect RPC and expire without being mined (~5 minutes), the payout order remains stuck in `PAYOUT_PENDING` status indefinitely. This blocks the affected transaction and requires manual intervention.

## Solution
New `processStuckPayoutOrders()` method in `PayoutService` that:
1. Finds `PAYOUT_PENDING` orders older than the configured timeout
2. Automatically retries them with the same nonce but current gas prices
3. The same nonce ensures the old (expired) transaction is replaced

## Changes
- `src/config/config.ts`: Added `payout.stuckPayoutTimeoutMinutes` config (default: 10 min)
- `src/subdomains/supporting/payout/services/payout.service.ts`: Added retry logic

## Test plan
- [ ] Deploy to staging
- [ ] Verify stuck orders are automatically retried after 10 minutes
- [ ] Verify logging shows retry attempts
- [ ] Verify successful retry completes the payout

### DIFF
--- a/src/config/config.ts
+++ b/src/config/config.ts
@@ -679,10 +679,6 @@ export class Configuration {
     ],
   };
 
-  payout = {
-    stuckPayoutTimeoutMinutes: +(process.env.PAYOUT_STUCK_TIMEOUT_MINUTES ?? 10),
-  };
-
   blockchain = {
     default: {
       user: process.env.NODE_USER,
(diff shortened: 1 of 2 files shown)


TaprootFreak:
Ready @davidleomay

PR #3002, opened 2026-01-20
feat: create RealUnit receipt for one transaction endpoint

xlamn:
#### Pre-Release

- [ ] Check migrations
  - Login/logout
  - Buy/sell payment request
  - KYC page

#### Post-Release

- Test basic user operations

### DIFF
--- a/src/shared/utils/pdf.util.ts
+++ b/src/shared/utils/pdf.util.ts
@@ -12,6 +12,11 @@ export enum PdfBrand {
   REALUNIT = 'REALUNIT',
 }
 
+export enum LogoSize {
+  SMALL = 'SMALL',
+  LARGE = 'LARGE',
+}
+
 export interface BalanceEntry {
   asset: Asset;
   balance: number;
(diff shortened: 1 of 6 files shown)


src/subdomains/supporting/payment/services/swiss-qr.service.ts
```diff
@@ -378,6 +357,18 @@ export class SwissQRService {
     } as Creditor;
   }
 
+  private realunitCreditor(): Creditor {
+    const realunitBank = Config.blockchain.realunit.bank;
+    return {
```
xlamn: This info is also hardcoded in https://github.com/DFXswiss/api/blob/9c1f4f57f0b9d82dde301b7e93d75b4f4c36387a/src/subdomains/supporting/realunit/realunit.service.ts#L251. We could think about putting these information in the config and .env.

src/shared/utils/pdf.util.ts
```diff
@@ -61,6 +62,45 @@ export class PdfUtil {
     pdf.translate(0, 30);
   }
 
+  static drawInvoiceLogo(pdf: InstanceType<typeof PDFDocument>, brand: PdfBrand = PdfBrand.DFX): void {
```
xlamn: this function and `drawLogo` are almost the same. They do some stuff differently so I was not sure if I can merge it that easily. What do you think @davidleomay?

davidleomay: The only difference is positioning and scale, right?
```TS
    pdf.translate(mm2pt(20), mm2pt(14));
    pdf.scale(0.15);
```
Maybe we can have a size parameter (small/large) then?

PR #3057, opened 2026-01-27
feat: split RealUnit register endpoint

xlamn:
- split register endpoint into register/email and register/complete
- register/email responsible to check if email is already in system or not
- register/complete responsible to fill out userdata and forwarding to Aktionariat

#### Pre-Release

- [ ] Check migrations
  - Login/logout
  - Buy/sell payment request
  - KYC page

#### Post-Release

- Test basic user operations

### DIFF
--- a/src/subdomains/supporting/realunit/realunit.service.ts
+++ b/src/subdomains/supporting/realunit/realunit.service.ts
@@ -359,10 +359,11 @@ export class RealUnitService {
         await this.userDataService.trySetUserMail(userData, dto.email);
       } catch (e) {
         if (e instanceof ConflictException) {
-          return RealUnitEmailRegistrationStatus.MERGE_REQUESTED;
-        } else {
-          throw e;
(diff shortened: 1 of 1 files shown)


src/subdomains/supporting/realunit/realunit.service.ts
```diff
@@ -341,6 +351,102 @@ export class RealUnitService {
...
+        const mergeUser = conflictUsers.find((u) => u.isMergePossibleWith(userData));
+        if (mergeUser) {
+          const mergeRequested = await this.accountMergeService.sendMergeRequest(mergeUser, userData, MergeReason.MAIL);
+          if (mergeRequested) {
+            return RealUnitEmailRegistrationStatus.MERGE_REQUESTED;
+          }
+        }
+        throw new ConflictException('Account already exists');
+      }
+    }
```
davidleomay: Do we really need to duplicate this from UserDataService? The check is also done in `trySetUserMail`

PR #3083, opened 2026-01-31
Add GET /support/kycFileList endpoint

TaprootFreak:
## Summary
- Add new compliance endpoint `GET /support/kycFileList`
- Returns all UserData entries where `kycFileId > 0`
- Response includes: kycFileId, id, amlAccountType, verifiedName

## Changes
- `user-data-support.dto.ts`: Add `KycFileListEntry` DTO
- `user-data.service.ts`: Add `getUserDatasWithKycFile()` method
- `support.service.ts`: Add `getKycFileList()` method
- `support.controller.ts`: Add `GET /support/kycFileList` endpoint (requires COMPLIANCE role)

## Test plan
- [ ] Call `GET /support/kycFileList` with valid COMPLIANCE token
- [ ] Verify response contains correct data structure
- [ ] Verify only entries with `kycFileId > 0` are returned

### DIFF
--- a/src/subdomains/generic/support/support.service.ts
+++ b/src/subdomains/generic/support/support.service.ts
@@ -107,10 +107,8 @@ export class SupportService {
     };
   }
 
-  async getKycFileList(): Promise<KycFileListEntry[]> {
-    const userDatas = await this.userDataService.getUserDatasWithKycFile();
-
-    return userDatas.map((u) => this.toKycFileListEntry(u));
+  getKycFileList(): Promise<KycFileListEntry[]> {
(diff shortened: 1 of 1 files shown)


src/subdomains/generic/support/support.service.ts
```diff
@@ -106,8 +107,23 @@ export class SupportService {
     };
   }
 
+  async getKycFileList(): Promise<KycFileListEntry[]> {
+    const userDatas = await this.userDataService.getUserDatasWithKycFile();
+
+    return userDatas.map((u) => this.toKycFileListEntry(u));
```
Yannick1712: Use `.then()`

Yannick1712 (approved):

PR #3198, opened 2026-02-16
fix: add Cardano to payment request broadcast hint

bernd2022:
## Summary
- Add `Blockchain.CARDANO` to the list of blockchains that receive the broadcast-style hint in payment request responses (alongside Monero, Zano, Solana, and Tron)

## Test plan
- [ ] Verify Cardano payment requests return the correct broadcast hint
- [ ] Verify other blockchains are unaffected


davidleomay (approved):
Fix formatting

PR #3213, opened 2026-02-18
fix: skip DFX trading limit for RealUnit transactions

TaprootFreak:
## Summary
- Skip the DFX 1000 CHF/month trading limit for REALU transactions, since Aktionariat manages its own limits
- Override `kycLimit` to unlimited for REALU in `getTxDetails()` — fixes `LIMIT_EXCEEDED` error and wrong `maxVolume`
- Skip `BANK_TRANSACTION_OR_VIDEO_MISSING` check for REALU in `getTxError()`

## Changes
- **1 file changed**: `src/subdomains/supporting/payment/services/transaction-helper.ts`
- Added `isRealUnitTransaction()` static helper method
- All other checks (AML, nationality, KYC level, trade approval, min/max specs) remain enforced

## Test plan
- [ ] `PUT /realunit/buy` with amount > 1000 CHF → should return `isValid: true`
- [ ] `PUT /realunit/buy/:id/confirm` → should succeed (no 400)
- [ ] Non-REALU transactions still respect the 1000 CHF limit for users < KYC 50
- [ ] AML and nationality checks still enforced for REALU

### DIFF
--- a/src/subdomains/supporting/payment/services/transaction-helper.ts
+++ b/src/subdomains/supporting/payment/services/transaction-helper.ts
@@ -291,8 +291,7 @@ export class TransactionHelper implements OnModuleInit {
     // get specs (CHF)
     const specs = this.getMinSpecs(from, to);
 
-    const { kycLimit: rawKycLimit, defaultLimit } = await this.getLimits(paymentMethodIn, paymentMethodOut, user);
-    const kycLimit = TransactionHelper.isRealUnitTransaction(from, to) ? Number.MAX_VALUE : rawKycLimit;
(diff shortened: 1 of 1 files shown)


src/subdomains/supporting/payment/services/transaction-helper.ts
```diff
@@ -1033,4 +1036,8 @@ export class TransactionHelper implements OnModuleInit {
 
     return currency;
   }
+
+  private static isRealUnitTransaction(from: Active, to: Active): boolean {
+    return (isAsset(from) && from.name === 'REALU') || (isAsset(to) && to.name === 'REALU');
+  }
 }
```
Yannick1712: Change this method as normal non static method and call it in `getLimits` method?

PR #3552, opened 2026-04-09
feat: add support dashboard API endpoints

bernd2022:
## Summary
- Add `MARKETING` role with proper hierarchy in role guard
- Add `GET support/issue/support/list` endpoint with filtering, pagination and full-text search (including message content via EXISTS subquery)
- Add `SupportIssueListDto`, `SupportIssueInternalTransactionMissingDataDto` and `GetSupportIssueListFilter`
- Add `clerk` field to issue detail response
- Make `message` optional in `CreateSupportMessageDto` (allows file-only messages)
- Remove mail requirement for issue creation (notification service already handles missing mail gracefully)
- Fix department filter to enforce role-based access (prevent cross-department access)
- Lower support user search guard from COMPLIANCE to SUPPORT

## Test plan
- [ ] Verify list endpoint returns correct issues per role/department
- [ ] Verify search works across name, UID, clerk and message content
- [ ] Verify file-only messages can be sent
- [ ] Verify issues can be created for users without email
- [ ] Run full test suite


src/shared/auth/role.guard.ts
```diff
@@ -13,12 +13,14 @@ class RoleGuardClass implements CanActivate {
...
       UserRole.COMPLIANCE,
+      UserRole.MARKETING,
       UserRole.REALUNIT,
     ],
     [UserRole.USER]: [UserRole.VIP, UserRole.BETA, UserRole.ADMIN, UserRole.SUPER_ADMIN, UserRole.CUSTODY],
     [UserRole.VIP]: [UserRole.ADMIN, UserRole.SUPER_ADMIN],
     [UserRole.BETA]: [UserRole.ADMIN, UserRole.SUPER_ADMIN],
-    [UserRole.SUPPORT]: [UserRole.COMPLIANCE, UserRole.ADMIN, UserRole.SUPER_ADMIN],
+    [UserRole.SUPPORT]: [UserRole.COMPLIANCE, UserRole.MARKETING, UserRole.ADMIN, UserRole.SUPER_ADMIN],
+    [UserRole.MARKETING]: [UserRole.ADMIN, UserRole.SUPER_ADMIN],
```
davidleomay, line 23: I think marketing should not be able to access support endpoints

src/subdomains/generic/support/support.controller.ts
```diff
@@ -38,7 +38,7 @@ export class SupportController {
   @Get()
   @ApiBearerAuth()
   @ApiExcludeEndpoint()
-  @UseGuards(AuthGuard(), RoleGuard(UserRole.COMPLIANCE), UserActiveGuard())
+  @UseGuards(AuthGuard(), RoleGuard(UserRole.SUPPORT), UserActiveGuard())
   async searchUserByKey(@Query() query: UserDataSupportQuery): Promise<UserDataSupportInfoResult> {
```
davidleomay, line 42: Do we need to hide some data for non-compliance users?

src/subdomains/supporting/support-issue/services/support-issue.service.ts
```diff
@@ -70,9 +76,6 @@ export class SupportIssueService {
   }
 
   async createIssueInternal(userData: UserData, dto: CreateSupportIssueDto): Promise<SupportIssueDto> {
-    // mail is required
-    if (!userData.mail) throw new BadRequestException('Mail is missing');
-
```
davidleomay, line 75: Why is this removed?

src/subdomains/supporting/support-issue/dto/support-issue-dto.mapper.ts
```diff
@@ -139,4 +163,20 @@ export class SupportIssueDtoMapper {
       limit: limitRequest.limit,
     };
   }
+
+  static mapTransactionMissingData(issue: SupportIssue): SupportIssueInternalTransactionMissingDataDto | undefined {
+    let info: { senderIban?: string; receiverIban?: string; date?: string } | undefined;
+    try {
+      info = issue.additionalInformation as typeof info;
+    } catch {
+      return undefined;
```
davidleomay: How can this throw an error?

src/subdomains/supporting/support-issue/services/support-issue.service.ts
```diff
@@ -210,6 +213,66 @@ export class SupportIssueService {
...
+
+    if (filter.state) where.state = filter.state;
+    if (filter.type) where.type = filter.type;
+
+    // server-side search: split query into terms, each term must match at least one field (AND between terms, OR between fields)
+    const terms = (filter.query ?? '')
+      .split(/\s+/)
+      .map((t) => t.trim())
+      .filter((t) => t.length > 0)
+      .slice(0, 10);
```
davidleomay, line 246: Do we use this server side search?

src/subdomains/supporting/support-issue/services/support-issue.service.ts
bernd2022: If a customer signs up using a wallet, they are not required to provide an email address. If this check is enabled, support will not be able to send a support message to the customer.

src/subdomains/generic/support/support.controller.ts
bernd2022: I checked with the compliance agent again. There are no concerns regarding the information displayed in the search field.

src/subdomains/supporting/support-issue/services/support-issue.service.ts
bernd2022: Yes, used by the support dashboard for the Canceled/Completed tabs (`services` repo, `support-dashboard.screen.tsx:83-92`). Those tabs load paged, so client-side filtering wouldn't find issues outside the current page — the query param hits `uid`, `clerk`, user names and message content, matching the search input placeholder.

src/subdomains/supporting/support-issue/services/support-issue.service.ts
davidleomay: The user is notified by mail when a new support message arrives. That's why this check is there 🤔

src/subdomains/supporting/support-issue/support-issue.controller.ts
```diff
@@ -89,7 +105,7 @@ export class SupportIssueController {
     @Param('id') id: string,
     @Body() dto: CreateSupportMessageDto,
   ): Promise<SupportMessageDto> {
-    return [UserRole.SUPPORT, UserRole.COMPLIANCE, UserRole.ADMIN].includes(jwt?.role)
+    return jwt?.role && [UserRole.SUPPORT, UserRole.COMPLIANCE, UserRole.ADMIN].includes(jwt.role)
```
davidleomay, line 108: Is Marketing missing here?

src/subdomains/supporting/support-issue/support-issue.controller.ts
```diff
@@ -62,6 +67,17 @@ export class SupportIssueController {
     return this.supportIssueService.getIssues(jwt?.account);
   }
 
+  @Get('support/list')
+  @ApiBearerAuth()
```
davidleomay, line 71: `  @Get('list')`?

src/subdomains/supporting/support-issue/support-issue.controller.ts
bernd2022: No, Marketing was intentionally removed here — consistent with your earlier comment on `role.guard.ts:23`. Since Marketing is no longer part of the SUPPORT role hierarchy, it can't reach this endpoint anyway.

Also worth noting: no current code path produces `Department.MARKETING` tickets, so the Marketing department is effectively unused today. If we want Marketing to manage their own tickets in the future, I'd propose doing that as a separate feature PR — happy to align on that, let me know what you prefer.

davidleomay (approved):

PR #3568, opened 2026-04-10
[NOTASK] set recommendationRef with user creation

Yannick1712:
#### Pre-Release

- [ ] Check migrations
  - Login/logout
  - Buy/sell payment request
  - KYC page

#### Post-Release

- Test basic user operations

### DIFF
--- a/src/subdomains/generic/user/models/auth/auth.service.ts
+++ b/src/subdomains/generic/user/models/auth/auth.service.ts
@@ -144,7 +144,12 @@ export class AuthService {
     userDataId?: number,
     userId?: number,
   ): Promise<AuthResponseDto> {
-    const userData = userDataId && (await this.userDataService.getUserData(userDataId, { users: true }));
+    const userData =
+      userDataId &&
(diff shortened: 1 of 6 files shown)


src/subdomains/generic/user/models/user-data/user-data.entity.ts
```diff
@@ -396,6 +397,9 @@ export class UserData extends IEntity {
   @OneToMany(() => SupportIssue, (supportIssue) => supportIssue.userData)
   supportIssues?: SupportIssue[];
 
+  @OneToMany(() => Recommendation, (recommendation) => recommendation.recommended)
+  recommendedRecommendations?: Recommendation[];
```
davidleomay, line 401: Maybe we could use `givenRecommendations` and `receivedRecommendations`?

src/subdomains/generic/user/models/auth/auth.service.ts
davidleomay, line 1: What about UserData merge, do we need to change something there?

PR #3576, opened 2026-04-15
feat: compliance improvements (limit request, stop transaction, chargeback)

bernd2022:
## Summary
- **Limit Request via support endpoint** (`bf5f6956a` → `ddfe3f858`): extend `CreateSupportIssueSupportDto` with optional `limitRequest` and route to COMPLIANCE department, mirroring the public createIssue endpoint.
- **Stop transaction endpoint** (`6da95caa2` → `723ee4c7a`): new `POST /transaction/admin/:id/stop` for COMPLIANCE users to halt a BuyCrypto transaction (sets `BuyCryptoStatus.STOPPED`). Mapper exposes the new `STOPPED` transaction state.
- **Chargeback endpoint** (`a0da59e98`): new `PUT /support/transaction/:id/chargeback` to release a refund for BuyCrypto / BuyFiat / unassigned BankTx, recording the agent identity (`Compliance/<Vorname>.<Nachname>`). Adds `GET /support/me`, exposes `chargebackAsset` on the transaction DTO and adds an optional `chargebackAmount` override on `TransactionRefundDto`.

## Companion PRs
- packages: DFXswiss/packages#145 (adds `TransactionState.STOPPED`)
- services: linked separately

## Test plan
- [ ] `npm run format:check` ✅
- [ ] `npm run lint` ✅
- [ ] `npm test` ✅ (66 suites, 927 tests passing)
- [ ] Manual: COMPLIANCE user can call `POST /transaction/admin/{id}/stop` and the transaction state becomes `Stopped`.
- [ ] Manual: COMPLIANCE user can call `GET /support/transaction/{id}/refund` followed by `PUT /support/transaction/{id}/chargeback` for BuyCrypto, BuyFiat and unassigned BankTx; `chargebackAllowedBy` records the agent.
- [ ] Manual: `GET /support/me` returns `{ id, firstname, surname }` of the calling COMPLIANCE user.
- [ ] Manual: COMPLIANCE user can submit a Limit Request via `POST /support/issue/...` with `limitRequest` payload; department resolves to COMPLIANCE.

### DIFF
--- a/src/subdomains/core/history/dto/transaction-refund.dto.ts
+++ b/src/subdomains/core/history/dto/transaction-refund.dto.ts
@@ -1,6 +1,6 @@
 import { ApiProperty, ApiPropertyOptional } from '@nestjs/swagger';
 import { Transform, Type } from 'class-transformer';
-import { IsNotEmpty, IsOptional, IsString, ValidateNested } from 'class-validator';
+import { IsNotEmpty, IsNumber, IsOptional, IsString, ValidateNested } from 'class-validator';
 import { Util } from 'src/shared/utils/util';
 
 export class CreditorDataDto {
@@ -48,6 +48,11 @@ export class TransactionRefundDto {
   @ValidateNested()
   @Type(() => CreditorDataDto)
   creditorData?: CreditorDataDto;
+
+  @ApiPropertyOptional({ description: 'Manual chargeback amount override' })
+  @IsOptional()
+  @IsNumber()
+  chargebackAmount?: number;
 }
 
 export class BankRefundDto extends TransactionRefundDto {
--- a/src/subdomains/core/history/dto/transaction-refund.dto.ts
+++ b/src/subdomains/core/history/dto/transaction-refund.dto.ts
@@ -48,7 +48,9 @@ export class TransactionRefundDto {
   @ValidateNested()
   @Type(() => CreditorDataDto)
   creditorData?: CreditorDataDto;
+}
 
+export class ChargebackRefundDto extends TransactionRefundDto {
   @ApiPropertyOptional({ description: 'Manual chargeback amount override' })
   @IsOptional()
   @IsNumber()
--- a/src/subdomains/generic/support/support.controller.ts
+++ b/src/subdomains/generic/support/support.controller.ts
@@ -12,11 +12,13 @@ import {
 } from '@nestjs/common';
 import { AuthGuard } from '@nestjs/passport';
 import { ApiBearerAuth, ApiExcludeEndpoint, ApiOkResponse } from '@nestjs/swagger';
+import { GetJwt } from 'src/shared/auth/get-jwt.decorator';
+import { JwtPayload } from 'src/shared/auth/jwt-payload.interface';
 import { RoleGuard } from 'src/shared/auth/role.guard';
 import { UserActiveGuard } from 'src/shared/auth/user-active.guard';
 import { UserRole } from 'src/shared/auth/user-role.enum';
 import { RefundDataDto } from 'src/subdomains/core/history/dto/refund-data.dto';
-import { BankRefundDto } from 'src/subdomains/core/history/dto/transaction-refund.dto';
+import { BankRefundDto, TransactionRefundDto } from 'src/subdomains/core/history/dto/transaction-refund.dto';
 import { GenerateOnboardingPdfDto } from './dto/onboarding-pdf.dto';
 import { TransactionListQuery } from './dto/transaction-list-query.dto';
 import {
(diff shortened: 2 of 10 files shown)


src/subdomains/core/history/dto/transaction-refund.dto.ts
```diff
@@ -48,6 +48,11 @@ export class TransactionRefundDto {
   @ValidateNested()
   @Type(() => CreditorDataDto)
   creditorData?: CreditorDataDto;
+
+  @ApiPropertyOptional({ description: 'Manual chargeback amount override' })
+  @IsOptional()
+  @IsNumber()
+  chargebackAmount?: number;
```
davidleomay, line 57: Is a bit dangerous when the user can override the chargeback amount?

src/subdomains/generic/support/support.controller.ts
```diff
@@ -43,6 +45,14 @@ export class SupportController {
     return this.supportService.searchUserDataByKey(query);
   }
 
+  @Get('me')
+  @ApiBearerAuth()
+  @ApiExcludeEndpoint()
+  @UseGuards(AuthGuard(), RoleGuard(UserRole.COMPLIANCE), UserActiveGuard())
+  async getSupportUserInfo(@GetJwt() jwt: JwtPayload): Promise<{ id: number; firstname: string; surname: string }> {
+    return this.supportService.getSupportUserInfo(jwt.account);
+  }
```
davidleomay: What is this used for?

src/subdomains/generic/support/support.controller.ts
```diff
@@ -139,4 +149,16 @@ export class SupportController {
...
+  @ApiBearerAuth()
+  @ApiExcludeEndpoint()
+  @UseGuards(AuthGuard(), RoleGuard(UserRole.COMPLIANCE), UserActiveGuard())
+  async chargebackTransaction(
+    @Param('id') id: string,
+    @Body() dto: TransactionRefundDto,
+    @GetJwt() jwt: JwtPayload,
+  ): Promise<void> {
+    await this.supportService.processChargebackRefund(+id, dto, jwt.account);
+  }
```
davidleomay: Can't we combine this into `@Put('transaction/:id/refund')`?

src/subdomains/supporting/support-issue/support-issue.controller.ts
```diff
@@ -57,7 +57,14 @@ export class SupportIssueController {
...
     @Body() dto: CreateSupportIssueSupportDto,
   ): Promise<SupportIssueDto> {
-    return this.supportIssueService.createIssue(+userDataId, dto);
+    const input: CreateSupportIssueSupportDto = {
+      ...dto,
+      department:
+        dto.type === SupportIssueType.VERIFICATION_CALL || dto.limitRequest
+          ? Department.COMPLIANCE
+          : Department.SUPPORT,
+    };
```
davidleomay, line 64: Maybe we should use the user role for this switch?

src/subdomains/core/history/dto/transaction-refund.dto.ts
bernd2022: That was a compliance requirement. Currently, the amount can be defined via the Google Sheet, so it should also be possible to do this via the compliance tool in the future.

Or is there a better solution for this?

src/subdomains/generic/support/support.controller.ts
bernd2022: Currently, the current employee must be selected manually in a few places in the Google Sheet.

This will no longer be necessary here, as the employee is identified through the login

src/subdomains/generic/support/support.controller.ts
bernd2022: I think keeping them separate is actually cleaner here. The two endpoints serve different operations with different semantics:

- **`PUT /refund`** handles returns of **unassigned bank transactions** — these are simple cases where money came in but couldn't be matched. The `BankRefundDto` enforces all creditor fields at the DTO level, which is correct since bank returns always need them.

- **`PUT /chargeback`** handles chargebacks for **BuyCrypto/BuyFiat transactions** that failed AML — a more complex flow with multiple refund paths (bank, crypto, card), optional `chargebackAmount` override, and compliance agent tracking via JWT.

Combining them into one endpoint would mean accepting the more permissive `TransactionRefundDto` for all cases, losing the compile-time validation of `BankRefundDto`. It would also make the endpoint responsible for too many different flows, making it harder to reason about.

The `GET /refund` endpoint already serves both cases (preparing refund data), which is fine since that's read-only. But for the write operations, the separation keeps the contracts explicit.

src/subdomains/core/history/dto/transaction-refund.dto.ts
davidleomay: Makes sense, but the "normal" (non-employee) user should not be able to override the chargeback amount for his transactions.

src/subdomains/generic/support/support.controller.ts
davidleomay: Does it make sense to pass the data through the frontend? We can just fetch the required data when the user calls API endpoints.

src/subdomains/generic/support/support.controller.ts
davidleomay: Hm, I'm not sure 🤣  for the user endpoints we also have only a single PUT endpoint, DTO validation can be done conditionally.
If we want to keep two endpoints, they should at least be named and use paths so one understands what they are doing. Both do refunds, one is just specific.

davidleomay (approved):

PR #3585, opened 2026-04-20
feat: require phone verification for users referred by non-trading accounts

TaprootFreak:
## Summary
- Adds a new AML check (`REFERRER_NO_TRADE_HISTORY`) that blocks transactions when the referrer has never completed a purchase (`buyVolume === 0`)
- Affected users are held in `PENDING` until phone verification is completed
- Trusted referrers (`isTrustedReferrer`) are exempt from this check
- Follows existing phone call status handling (failed/rejected calls result in `FAIL`)

## Context
Referral abuse case where an account registered via an `autoTradeApproval` wallet, obtained `tradeApprovalDate` automatically, and then used mail invitations to approve referred users — bypassing the trade approval gate without ever making a purchase.

## Test plan
- [ ] Verify transactions from users with a non-trading referrer are held in `PENDING`
- [ ] Verify `phoneCallCheckDate` clears the block after successful phone verification
- [ ] Verify trusted referrers are not affected
- [ ] Verify users without a referrer (`refUser === undefined`) are not affected
- [ ] Verify referrers with `buyVolume > 0` do not trigger the check


src/subdomains/core/aml/services/aml-helper.service.ts
```diff
@@ -80,6 +80,19 @@ export class AmlHelperService {
...
+      !entity.userData.phoneCallCheckDate &&
+      !refUser.userData.isTrustedReferrer
+    )
+      errors.push(
+        entity.userData.phoneCallStatus === PhoneCallStatus.FAILED
+          ? AmlError.USER_DATA_FAILED_CALL
+          : entity.userData.phoneCallStatus === PhoneCallStatus.USER_REJECTED && !entity.userData.phoneCallAccepted
+            ? AmlError.USER_DATA_REJECTED_CALL
+            : AmlError.REFERRER_NO_TRADE_HISTORY,
+      );
```
davidleomay, line 97: `refUser` ist der Ref User (wird über den Ref bestimmt) und nicht der Recommender.

src/subdomains/core/aml/services/aml.service.ts
```diff
@@ -125,6 +129,8 @@ export class AmlService {
     let bankData = await this.getBankData(entity);
     const refUser =
       entity.user.usedRef !== Config.defaultRef ? await this.userService.getRefUser(entity.user.usedRef) : undefined;
+    const recommendations = await this.recommendationService.getRecommendationsByRecommendedId(entity.userData.id);
+    const recommender = recommendations.find((r) => r.recommender)?.recommender;
```
davidleomay: Find the confirmed recommendation?

davidleomay (approved):

PR #3593, opened 2026-04-21
feat: support dashboard counts endpoint, configurable clerks, message stats single query

bernd2022:
## Summary
- **New endpoints**: `GET /support/issue/counts` (single `GROUP BY` query replaces 3 list calls) and `GET /support/issue/clerks` (reads from `supportClerks` setting, falls back to `['Support']`).
- **Setting**: registered `supportClerks` (`string[]`) in the schema registry so clerk names are configurable instead of hardcoded in the frontend.
- **Performance**: `getMessageStats` now uses a single correlated-subquery query instead of two round-trips.
- **DTO cleanup**: dropped single `state` filter in favor of `states` (comma-separated), added reusable `@StringToArray()` transform decorator, Swagger now advertises the enum with `isArray`.
- **Bug fix**: restored last-writer-wins clerk assignment on message creation (previous change left `AutoResponder` stuck as clerk).

## Test plan
- [ ] `GET /support/issue/counts` returns `{ Created, Pending, OnHold, Completed, Canceled }` counts honoring role-based department filter
- [ ] `GET /support/issue/clerks` returns configured list; `['Support']` when setting is missing or empty
- [ ] `GET /support/issue/list?states=OnHold,Canceled` filters correctly
- [ ] Sending a support reply updates `issue.clerk` to the message author
- [ ] Full test suite passes (`npm test`)

### DIFF
--- a/src/subdomains/supporting/support-issue/services/support-issue.service.ts
+++ b/src/subdomains/supporting/support-issue/services/support-issue.service.ts
@@ -33,7 +33,7 @@ import {
 import { UpdateSupportIssueDto } from '../dto/update-support-issue.dto';
 import { SupportIssue } from '../entities/support-issue.entity';
 import { AutoResponder, CustomerAuthor, SupportMessage } from '../entities/support-message.entity';
-import { Department } from '../enums/department.enum';
+import { RoleDepartmentMap } from '../enums/department.enum';
 import { SupportIssueInternalState, SupportIssueReason, SupportIssueType } from '../enums/support-issue.enum';
 import { SupportLogType } from '../enums/support-log.enum';
 import { SupportIssueRepository } from '../repositories/support-issue.repository';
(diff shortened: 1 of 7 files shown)


src/subdomains/supporting/support-issue/services/support-issue.service.ts
```diff
@@ -55,8 +56,40 @@ export class SupportIssueService {
...
+    const qb = this.supportIssueRepo
+      .createQueryBuilder('issue')
+      .select('issue.state', 'state')
+      .addSelect('COUNT(*)', 'count')
+      .groupBy('issue.state');
+
+    const roleDepartmentMap: Partial<Record<UserRole, Department>> = {
+      [UserRole.SUPPORT]: Department.SUPPORT,
+      [UserRole.COMPLIANCE]: Department.COMPLIANCE,
+    };
```
Yannick1712: I would move that to an mapper variable outside the method

davidleomay: Maybe we can already reuse it somewhere?

src/subdomains/supporting/support-issue/services/support-issue.service.ts
```diff
@@ -273,7 +322,42 @@ export class SupportIssueService {
...
+        .select('m.issueId', 'issueId')
+        .addSelect('COUNT(*)', 'count')
+        .addSelect(
+          `(SELECT TOP 1 m2.created FROM support_message m2 WHERE m2.issueId = m.issueId ORDER BY m2.id DESC)`,
+          'lastDate',
+        )
+        .addSelect(
+          `(SELECT TOP 1 m2.author FROM support_message m2 WHERE m2.issueId = m.issueId ORDER BY m2.id DESC)`,
+          'lastAuthor',
+        )
```
davidleomay, line 348: Can we use subquery instead of raw SQL?

davidleomay (approved):

PR #3599, opened 2026-04-22
feat: compliance pending reviews endpoint

bernd2022:
## Summary
- Adds `GET /support/pending-reviews` returning a summary of all `kyc_step` and `bank_data` entries in `ManualReview` or `InternalReview`, grouped by type and name with per-status counts.
- Adds `GET /support/pending-reviews/items?type=&name=&status=` returning the affected users for a given review category.
- Both endpoints mirror the auth/role guards of the existing `pending-onboardings` endpoint (Compliance role required).
- Refactors the existing name-fallback logic in `getPendingOnboardings` into a reusable helper to avoid drift.

## Context
Today the compliance team relies on an external Google Sheet (`NutzerMonitoring`) to see pending manual checks. The new endpoints expose the same data from the internal database, enabling the services frontend to render a `Pending Reviews` section on the compliance landing page (see companion PR in the services repo).

The implementation covers a superset of the Google Sheet scope — in addition to all explicitly tracked step types, it includes `NameChange`, `AddressChange`, `PhoneChange`, and any other `KycStepName` that happens to be in `ManualReview`/`InternalReview`.

## Test plan
- [ ] `/support/pending-reviews` returns one entry per `(type, name)` with `manualReview` and `internalReview` counts matching the DB.
- [ ] `/support/pending-reviews/items?type=KycStep&name=Ident&status=ManualReview` returns the expected users.
- [ ] `/support/pending-reviews/items?type=BankData&status=ManualReview` returns the expected bank data entries.
- [ ] Unauthenticated requests return 401; non-compliance roles return 403.
- [ ] Merged users are excluded.
- [ ] Existing `/support/pending-onboardings` endpoint is unaffected.

### DIFF
--- a/src/subdomains/generic/kyc/services/kyc.service.ts
+++ b/src/subdomains/generic/kyc/services/kyc.service.ts
@@ -1893,13 +1893,11 @@ export class KycService {
   }
 
   async getPendingReviewSteps(name: KycStepName, status: ReviewStatus): Promise<KycStep[]> {
-    return this.kycStepRepo
-      .createQueryBuilder('step')
-      .innerJoinAndSelect('step.userData', 'userData')
-      .where('step.name = :name', { name })
-      .andWhere('step.status = :status', { status })
-      .andWhere('userData.status != :mergedStatus', { mergedStatus: UserDataStatus.MERGED })
-      .orderBy('step.updated', 'ASC')
-      .getMany();
+    return this.kycStepRepo.find({
+      where: { name, status, userData: { status: Not(UserDataStatus.MERGED) } },
+      relations: { userData: true },
+      loadEagerRelations: false,
+      order: { updated: 'ASC' },
+    });
   }
 }
--- a/src/subdomains/generic/kyc/services/kyc.service.ts
+++ b/src/subdomains/generic/kyc/services/kyc.service.ts
@@ -1874,4 +1874,32 @@ export class KycService {
       relations: { userData: true },
     });
   }
+
+  // --- Pending Review Queries ---
+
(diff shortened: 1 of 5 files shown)


src/subdomains/generic/kyc/services/kyc.service.ts
```diff
@@ -1874,4 +1874,32 @@ export class KycService {
...
+
+  async getPendingReviewSteps(name: KycStepName, status: ReviewStatus): Promise<KycStep[]> {
+    return this.kycStepRepo
+      .createQueryBuilder('step')
+      .innerJoinAndSelect('step.userData', 'userData')
+      .where('step.name = :name', { name })
+      .andWhere('step.status = :status', { status })
+      .andWhere('userData.status != :mergedStatus', { mergedStatus: UserDataStatus.MERGED })
+      .orderBy('step.updated', 'ASC')
+      .getMany();
```
davidleomay: Can be done w/o query builder?

src/subdomains/generic/user/models/bank-data/bank-data.service.ts
```diff
@@ -465,4 +465,31 @@ export class BankDataService {
...
+  }
+
+  async getPendingReviewList(status: ReviewStatus): Promise<BankData[]> {
+    return this.bankDataRepo
+      .createQueryBuilder('bankData')
+      .innerJoinAndSelect('bankData.userData', 'userData')
+      .where('bankData.status = :status', { status })
+      .andWhere('userData.status != :mergedStatus', { mergedStatus: UserDataStatus.MERGED })
+      .orderBy('bankData.updated', 'ASC')
+      .getMany();
```
davidleomay: Can be done w/o query builder?

davidleomay (approved):

PR #3634, opened 2026-04-28
fix: chunk support issue message stats query to avoid SQL parameter limit

bernd2022:
## Summary
- `GET /v1/support/issue/list` failed with `The incoming request has too many parameters. The server supports a maximum of 2100 parameters` when more than ~2100 issues matched the filter.
- Root cause: `getMessageStats` passed all issue IDs into a single `IN (:...ids)` clause; TypeORM emits one parameter per ID and SQL Server rejects the query above 2100.
- Fix: chunk the message stats query in batches of 1000 IDs and merge the results into a single map. No API change, identical result set per issue.

## Test plan
- [x] `npx prettier --check` on the changed file
- [x] `npx eslint` on the changed file
- [x] `npm test` (927 passed, 0 failed)
- [ ] Verify `/v1/support/issue/list` succeeds on a dataset with >2100 matching issues

### DIFF
--- a/src/subdomains/supporting/support-issue/services/support-issue.service.ts
+++ b/src/subdomains/supporting/support-issue/services/support-issue.service.ts
@@ -321,41 +321,52 @@ export class SupportIssueService {
   ): Promise<Map<number, { count: number; lastDate?: Date; lastAuthor?: string }>> {
     if (issueIds.length === 0) return new Map();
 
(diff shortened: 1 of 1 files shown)


davidleomay (commented):
Could we use Util.doInBatches?

davidleomay (approved):

PR #3656, opened 2026-04-30
[NOTASK] phoneCall auto aml reset

Yannick1712:
#### Pre-Release

- [ ] Check migrations
  - Login/logout
  - Buy/sell payment request
  - KYC page

#### Post-Release

- Test basic user operations


davidleomay (approved):

src/subdomains/generic/user/models/user-data/user-data.service.ts
davidleomay, line 1: It would be cleaner to only have something like a UserData phone call completed event. The services would then listen to it and check themself, if they have a transaction to reset the AML check. WDYT?

src/subdomains/core/buy-crypto/process/services/buy-crypto.service.ts
```diff
@@ -101,8 +104,30 @@ export class BuyCryptoService {
...
+
+  async checkAmlResetTx(userData: UserData): Promise<void> {
+    const entities = await this.buyCryptoRepo.findBy({
+      transaction: { userData: { id: userData.id } },
+    });
+
+    for (const entity of entities) {
+      if (
+        Util.daysDiff(entity.created) > 90 &&
+        entity.amlCheck === CheckStatus.FAIL &&
```
davidleomay: `< 90`?

davidleomay (approved):

PR #3677, opened 2026-05-05
feat(payment-link): enable ADA payments on Cardano

joshuakrueger-dfx:
## Summary
- Sets `paymentEnabled = 1` on `Cardano/ADA` asset so it surfaces in payment-link `transferAmounts`
- Required for urble (and other Cardano-supporting wallets) to accept ADA at merchant payment pages like SPAR

## Test plan
- [ ] Migration runs without error
- [ ] `GET /v1/paymentLink/payment?route=SPAR&...` returns Cardano in `transferAmounts`
- [ ] urble wallet detail on /pl shows ADA as payment option

### DIFF
--- a/migration/1777989917652-EnableCardanoPayment.js
+++ b/migration/1777989917652-EnableCardanoPayment.js
@@ -1,0 +1,11 @@
+module.exports = class EnableCardanoPayment1777989917652 {
+  name = 'EnableCardanoPayment1777989917652';
+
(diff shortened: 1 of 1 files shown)


joshuakrueger-dfx:
Not needed — Cardano/ADA paymentEnabled is already true on prod; dev mismatch is intentional or out of scope here.

PR #3697, opened 2026-05-12
fix(payout): recover EVM payouts from out-of-gas reverts

TaprootFreak:
## Summary
Closes the failure mode where an EVM payout TX runs out of gas on-chain and the order stays in `PayoutPending` indefinitely.

## Root cause

Two defects had to coincide:

1. **Failure swallowed.** `isTxComplete` threw on reverted receipts. The outer `try/catch` in `PayoutService.checkPayoutCompletion` logged the error every 30 s but left the order in `PayoutPending`.
2. **No retry path for mined reverts.** `canRetryFailedPayout` only handles mempool-expired TXs (`isTxExpired`); a mined receipt with `status=0` is not expired and thus not retryable.

The underlying race itself — gas estimate vs execution diverging because of an EIP-2200 SSTORE swing on the recipient's balance slot — is rare in practice; the failure mode that needs fixing is the stuck order, not the rare gas estimate miss.

## Fix

- **`getPayoutCompletionData`** — returns a tri-state `PayoutTxStatus` (`pending` / `complete` / `failed` with `isOutOfGas`) read directly from the receipt instead of relying on the throwing `isTxComplete`.
- **`EvmStrategy.checkPayoutCompletionData`** — handles `failed` explicitly:
  - `isOutOfGas` → `rollbackPayout()` → next cron pass retries with a fresh nonce and a re-estimated gas limit.
  - non-OOG revert → `designatePayout()` → `processFailedOrders` already mails and moves the order to `PayoutUncertain`.
- **`PayoutOrder.rollbackPayout()`** — new entity method symmetric to `pendingPayout(txId)`: clears `payoutTxId` and resets status to `PreparationConfirmed`. Unit test added.
- **`sendToken`** now passes the real transfer amount through to `getTokenGasLimitForContact` (consistent with `prepareTxData`).
- Redundant `getPayoutCompletionData` overrides in `base-{coin,token}`, `gnosis-{coin,token}`, `optimism-{coin,token}`, `polygon-{coin,token}` strategies removed — all 18 EVM strategies now consistently use the parent implementation.
- `PayoutTxStatus` collocated with `PayoutRequest` / `FeeResult` in `payout/interfaces/index.ts`.

No DB migration, no new status, no new entity columns, no change to gas estimation values — everything routes through existing entity methods (`rollbackPayout`, `designatePayout`, `pendingInvestigation`) and the existing `processFailedOrders` mail+investigation path.

## Test plan
- [ ] CI green (format / type-check / lint / jest)
- [ ] Manual: trigger an EVM payout that previously failed, verify auto-rollback + retry produces a successful TX with a new nonce
- [ ] Manual: simulate a non-OOG revert (e.g. paused token), verify order ends in `PayoutUncertain` with mail

### DIFF
--- a/src/subdomains/supporting/payout/strategies/payout/impl/base/evm.strategy.ts
+++ b/src/subdomains/supporting/payout/strategies/payout/impl/base/evm.strategy.ts
@@ -69,7 +69,7 @@ export abstract class EvmStrategy extends PayoutStrategy {
           order.designatePayout();
           await this.payoutOrderRepo.save(order);
         } else if (await this.canRetryFailedPayout(order, status)) {
-          if (status.state === 'failed') {
+          if (status.state === 'failed' && status.isOutOfGas) {
             // OOG: free the spent nonce so the retry gets a fresh one
             this.logger.warn(
               `Payout order ${order.id} failed with out-of-gas (tx ${order.payoutTxId}), retrying with fresh nonce`,
--- a/src/subdomains/supporting/payout/strategies/payout/impl/base/evm.strategy.ts
+++ b/src/subdomains/supporting/payout/strategies/payout/impl/base/evm.strategy.ts
@@ -4,11 +4,11 @@ import { DfxLogger } from 'src/shared/services/dfx-logger';
 import { DisabledProcess, Process } from 'src/shared/services/process.service';
 import { AsyncCache, CacheItemResetPeriod } from 'src/shared/utils/async-cache';
 import { Util } from 'src/shared/utils/util';
-import { FeeResult } from 'src/subdomains/supporting/payout/interfaces';
+import { FeeResult, PayoutTxStatus } from 'src/subdomains/supporting/payout/interfaces';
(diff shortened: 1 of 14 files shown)


src/subdomains/supporting/payout/strategies/payout/impl/base/evm.strategy.ts
davidleomay: Can the `status.isOutOfGas` be combined with `canRetryFailedPayout`?

davidleomay (approved):

src/subdomains/supporting/payout/strategies/payout/impl/base/evm.strategy.ts
TaprootFreak: Done in a09c67c41 — both recoverable cases (OOG-mined and mempool-expired) now route through `canRetryFailedPayout`. OOG keeps its rollback-nonce step in the caller since it's the only divergence (spent nonce vs reusable one for expired).

davidleomay (approved):

PR #3725, opened 2026-05-19
perf(dashboard): optimize financial log queries (load 5-15s -> <500ms)

TaprootFreak:
## Summary

The Financial Overview dashboard page took 5-15s to load because the backend re-parsed the entire ``nvarchar(MAX)`` ``message`` JSON for every row in an un-indexed ``log`` table (~525k rows/year, cron writes every minute), shipped the full uncompressed payload, and had no caching layer.

This PR addresses all four hot spots without changing endpoint contracts.

## Changes

**Fix #1 - Composite index** (``migration/1779221816705-AddLogFinancialIndex.js``)
- ``IDX_7765c3f5f663a0c6d250d28255 (subsystem, severity, created)`` on ``log``
- Covers ``getLatestFinancialLog``, ``getFinancialLogs``, ``getFinancialChangesLogs`` and the per-minute ``maxEntity`` lookup from the cron
- Deterministic IDX name generated via the same SHA1 scheme TypeORM uses, so future ``npm run migration`` runs see no drift

**Fix #2 - Denormalised aggregate columns** (``migration/1779221847925-AddLogFinancialAggregates.js``)
- Five new nullable columns on ``log``: ``totalBalanceChf``, ``plusBalanceChf``, ``minusBalanceChf``, ``btcPriceChf``, ``balancesByTypeJson``
- ``log-job.service.ts`` populates them when writing each ``FinancialDataLog`` row
- ``dashboard-financial.service.ts`` ``mapLogToEntry()`` uses these as a fast path and only ``JSON.parse``s the nvarchar(MAX) ``message`` for legacy rows pre-migration
- ``getLatestBalance()`` still parses the full JSON (it needs the per-asset breakdown for the live cards) but is a single row, not thousands

**Fix #3 - DB-side sampling for sub-week ranges** (``log.repository.ts``)
- New ``getSampleIntervalMinutes(from, dailySample)`` helper: ``null`` for 24h (full per-minute resolution, ~1440 rows) and beyond 1 week, ``5`` for 3-day/1-week ranges (cuts 4320 rows to ~864)
- ``getFinancialLogs`` / ``getFinancialChangesLogs`` share a single ``getSampledFinancialLogs`` impl
- Bucketing uses ``MAX(id) GROUP BY DATEADD(MINUTE, (DATEDIFF(MINUTE, 0, created) / N) * N, 0)`` to mirror the existing ``dailySample`` pattern

**Fix #4 - gzip compression + service-layer caching** (``main.ts`` + ``dashboard-financial.service.ts``)
- ``app.use(compression())`` so JSON payloads are gzipped
- 30s ``AsyncCache`` on ``getLatestBalance``, ``getLatestFinancialChanges``, ``getFinancialLog``, ``getFinancialChanges`` (cron writes every minute, so a 30s TTL never serves data older than the next tick on average)
- Cache keys include ``from`` and ``dailySample`` query params; no shared state across requests with different parameters

## Test plan

- [x] ``npm run lint``
- [x] ``npm run format:check``
- [x] ``npm run build``
- [x] ``npm test`` — 947 passed (was 938), 0 failed. 9 new tests in ``log.repository.spec.ts`` (5) and ``dashboard-financial.service.spec.ts`` (4) cover the bucketing helper, the fast-path/legacy-fallback in ``mapLogToEntry``, and the cache TTL + cache-key behaviour
- [ ] ``npm run migration:run`` skipped locally — no MSSQL instance available; both migrations are plain ``ALTER TABLE`` / ``CREATE INDEX`` and will run on the DEV deploy
- [ ] Smoke-test the four dashboard endpoints on DEV after deploy

## Migration Notes

- Both migrations are **non-breaking**: the index is additive, and the new columns are nullable so existing rows keep working via the ``JSON.parse(message)`` fallback in the service
- The cron will start populating the new columns automatically on the first tick after deploy; backfilling old rows is **not** required
- Cache TTL is conservative (30s); the underlying cron writes every minute, so the maximum staleness an admin sees is ~30s on average and 60s worst-case — same order of magnitude as the data source itself


src/subdomains/supporting/log/log.repository.ts
davidleomay, line 1: Adapt to PSQL

migration/1779221816705-AddLogFinancialIndex.js
davidleomay, line 1: Adapt to PSQL

migration/1779221847925-AddLogFinancialAggregates.js
davidleomay, line 1: Adapt to PSQL

PR #3727, opened 2026-05-20
feat(auth): add Monitoring role with read access to financial dashboard

TaprootFreak:
## Summary

- Adds a new `MONITORING` user role with read-only access to the financial dashboard endpoints.
- Hierarchy: `ADMIN` ⊃ `COMPLIANCE` ⊃ `MONITORING` — all three roles are explicitly listed on the affected guards so each can hit the endpoints without changing the existing `additionalRoles` hierarchy map.
- Extends `RoleGuard` to accept multiple entry roles (variadic, backward-compatible — existing single-arg call sites are unaffected).

## Changes

- `src/shared/auth/user-role.enum.ts`: add `MONITORING = 'Monitoring'` (positioned after `MARKETING`, before service / external roles).
- `src/shared/auth/role.guard.ts`: make `RoleGuard(...entryRoles)` variadic; `canActivate` returns true if the request role matches (or is in the `additionalRoles` of) any of the supplied entry roles.
- `src/subdomains/supporting/dashboard/dashboard-financial.controller.ts`: widen guards on all four endpoints (`GET /dashboard/financial/log`, `/latest`, `/changes/latest`, `/ref-recipients`, `/changes`) from `RoleGuard(ADMIN)` to `RoleGuard(ADMIN, COMPLIANCE, MONITORING)`.

## Test plan

- [x] `npm run lint`
- [x] `npm run format` (no diff)
- [x] `npm run type-check`
- [x] `npm run build`
- [x] `npm test` (938 passed, 106 skipped — unchanged baseline)
- [ ] Manual: assign a user the `Monitoring` role and verify they can hit the five financial dashboard endpoints; verify they cannot hit other admin-only endpoints.

## Migration notes

None — additive enum value + guard widening only. No DB / schema changes.

## Coordination

Touches a different file than the in-flight PR #3725 (`perf/dashboard-financial-log-optimization` modifies the service, this PR modifies the controller + auth) — no merge conflicts expected.

### DIFF
--- a/src/shared/auth/role.guard.ts
+++ b/src/shared/auth/role.guard.ts
@@ -39,14 +39,16 @@ class RoleGuardClass implements CanActivate {
     [UserRole.CLIENT_COMPANY]: [UserRole.KYC_CLIENT_COMPANY],
   };
 
-  constructor(private readonly entryRole: UserRole) {}
+  constructor(private readonly entryRoles: UserRole[]) {}
 
   canActivate(context: ExecutionContext): boolean {
     const userRole = context.switchToHttp().getRequest().user?.role;
-    return this.entryRole === userRole || this.additionalRoles[this.entryRole]?.includes(userRole);
+    return this.entryRoles.some(
+      (entryRole) => entryRole === userRole || this.additionalRoles[entryRole]?.includes(userRole),
+    );
   }
 }
 
-export function RoleGuard(entryRole: UserRole): RoleGuardClass {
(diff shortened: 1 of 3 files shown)


src/shared/auth/role.guard.ts
davidleomay, line 1: Add ` [UserRole.MONITORING]: [UserRole.ADMIN, UserRole.SUPER_ADMIN],` to additionalRoles -> can remove admin on dashboard controller endpoints

PR #3731, opened 2026-05-21
fix(realunit): make register/wallet idempotent on signature match

TaprootFreak:
## Summary

Surface a real client retry-stuck bug spotted while reviewing [DFXswiss/realunit-app#466](https://github.com/DFXswiss/realunit-app/pull/466) (KYC merge flow). \`POST /v1/realunit/register/wallet\` (and the same-wallet branch of \`POST /v1/realunit/register/complete\`) was non-idempotent: if the first call succeeded server-side but the response was lost (mobile network drop is the realistic case), the client's automatic retry with the same EIP-712 signature got a hard \`400 Bad Request\` — "RealUnit registration already exists for this wallet" — with no recovery path. The user ends up stuck on a failure snackbar even though the registration is in fact complete.

## Fix

Replace the hard throw with a signature-aware idempotency check in \`RealUnitService\` (new private helper \`idempotentRegistrationResult\`):

- **Same wallet + same signature** → return the existing registration's status. \`COMPLETED\` if the existing \`KycStep\` is completed (forward to Aktionariat succeeded on the first attempt), \`FORWARDING_FAILED\` otherwise. **No new \`KycStep\`, no re-forward.**
- **Same wallet + different signature** → keep the 400, but with a more precise message: \`"RealUnit registration already exists for this wallet with a different signature"\`. A fresh sign over the same wallet means the client produced a new payload — that is a real conflict, not a retry.

Signature comparison is **case-insensitive** (\`Util.equalsIgnoreCase\`): EIP-712 signatures are 0x-prefixed hex strings and lower/upper case representations are semantically identical. Matches the convention already used for wallet-address comparisons in \`findRegistrationStep\`.

Both endpoints share the helper; \`completeRegistration\` is refactored to call \`findRegistrationStep\` directly instead of the boolean-only \`hasRegistrationForWallet\`, so it has access to the \`KycStep\` for the idempotency check. \`hasRegistrationForWallet\` itself is preserved (still used by \`realunit.controller.ts\` and two other call-sites).

The helper takes \`userData\` as an explicit parameter rather than reading \`step.userData\` so the call does not depend on TypeORM back-populating the inverse relation on steps loaded via \`userData.kycSteps\`. Each idempotent hit is logged at \`info\` level (\`userData.id\`, \`kycStep.id\`, resulting status) so production retry frequency is observable and stuck-user reports can be correlated to a specific step.

## Tests

\`describe('completeRegistrationForWalletAddress (idempotency)', ...)\` block in \`realunit.service.spec.ts\` covers all branches:

1. Same signature + completed step → returns \`COMPLETED\`, no \`createCustomKycStep\` call
2. Same signature + non-completed step → returns \`FORWARDING_FAILED\`, no \`createCustomKycStep\` call
3. Same signature, different case (stored upper, incoming lower) → returns \`COMPLETED\`, no \`createCustomKycStep\` call
4. Different signature → throws \`BadRequestException\`, no \`createCustomKycStep\` call

Test setup expands the \`UserService\` and \`KycService\` mock providers (previously empty objects) with the methods needed by these tests. The \`Util.equalsIgnoreCase\` jest mock was missing from the existing setup — added here too, since both \`findRegistrationStep\` and the new signature comparison rely on it.

## Local verification

- \`npm install\` ✅
- \`npm run format\` — no changes to touched files
- \`npm run lint\` ✅
- \`npm run type-check\` ✅
- \`npm test\` — 14/14 in \`realunit.service.spec.ts\` ✅

## Test plan

- [ ] **Manual reproduction of the original stuck state** (against DEV): trigger \`/v1/realunit/register/wallet\` twice with identical body — second call should now return \`201 { "status": "completed" }\` instead of \`400\`.
- [ ] **Signature-mismatch protection**: second call with same wallet but altered signature → still \`400\`, new message ("…with a different signature").
- [ ] **Unrelated flows unchanged**: new-customer registration (no existing step) still creates a \`KycStep\` and forwards as before.

## Known limitations

- **UTC midnight retry**: the realunit-app client signs over \`registrationDate = today\` (in UTC). A retry that crosses the UTC date boundary therefore re-signs against a different message and lands in the "different signature" branch (400). In the real-world failure mode (snackbar fires, user taps "retry" seconds later) this never triggers, but if a retry happens minutes after the rollover the client will still need to surface a sensible error. A future iteration could relax the match to wallet+userData once the forward has reached \`COMPLETED\`, but that loosens the conflict semantics deliberately defined here.

## Out of scope / follow-ups

- The realunit-app KYC merge flow already surfaces these failures correctly after [DFXswiss/realunit-app#466](https://github.com/DFXswiss/realunit-app/pull/466) — this PR closes the backend loop so legitimate retries succeed instead of looping in the failure path.
- \`completeRegistrationForWalletAddress\` still does not call \`validateRegistrationDto\` (only the signature is verified). Out of scope here, but worth a separate look — \`registrationDate\` and birthday/age validation are skipped on this path.

### DIFF
--- a/src/subdomains/supporting/realunit/realunit.service.ts
+++ b/src/subdomains/supporting/realunit/realunit.service.ts
@@ -604,9 +604,9 @@ export class RealUnitService {
       throw new BadRequestException('Email does not match registered email');
     }
 
-    const existingForWallet = this.findRegistrationStep(userData, dto.walletAddress);
-    if (existingForWallet.isForCurrentWallet) {
-      return this.idempotentRegistrationResult(existingForWallet.step!, dto.signature);
+    const { step: existingStep, isForCurrentWallet } = this.findRegistrationStep(userData, dto.walletAddress);
+    if (isForCurrentWallet) {
+      return this.idempotentRegistrationResult(userData, existingStep!, dto.signature);
     }
 
     // validate personal data
@@ -668,7 +668,7 @@ export class RealUnitService {
     const { step: registrationStep, isForCurrentWallet } = this.findRegistrationStep(userData, dto.walletAddress);
 
     if (isForCurrentWallet) {
-      return this.idempotentRegistrationResult(registrationStep!, dto.signature);
+      return this.idempotentRegistrationResult(userData, registrationStep!, dto.signature);
     }
 
     if (!registrationStep) {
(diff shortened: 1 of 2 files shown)


src/subdomains/supporting/realunit/realunit.service.ts
```diff
@@ -890,6 +891,40 @@ export class RealUnitService {
...
+   */
+  private idempotentRegistrationResult(
+    userData: UserData,
+    step: KycStep,
+    incomingSignature: string,
+  ): RealUnitRegistrationStatus {
+    const existingData = step.getResult<RealUnitRegistrationDto>();
+    if (existingData?.signature !== incomingSignature) {
+      throw new BadRequestException('RealUnit registration already exists for this wallet with a different signature');
+    }
```
davidleomay: Use case insensitive check? `Util.equalsIgnoreCase`

davidleomay (approved):

PR #3732, opened 2026-05-21
feat(kyc): expose decision fields (isRequired, processStatus) so clients can stop re-implementing routing locally

Blume1977:
## Summary

The realunit-app's `KycCubit` was rebuilding the API's own KYC routing rule client-side — a hardcoded `_requiredStepNames` set and an `actionableStatuses` set. Keeping that copy in sync with the backend is fragile; the 2026-05-21 incident (user_data 338759) is a symptom of it drifting.

This PR surfaces the two signals the app needs directly in the KYC DTOs so it can drop its own copy of the rule. Wave 2.1 of [DFXswiss/realunit-app `docs/api-authority-plan.md`](https://github.com/DFXswiss/realunit-app/blob/develop/docs/api-authority-plan.md); the companion app-side PR consumes these fields and removes the local logic.

### `KycStepDto.isRequired: boolean`

Populated from `requiredKycSteps(userData)` at mapping time. Clients drop their own duplicate sets and iterate `kycSteps.filter(s => s.isRequired)` instead. The required-step list stays in `kyc.enum.ts` — the DTO just exposes the per-user verdict.

### `KycLevelDto.processStatus: KycProcessStatus`

High-level KYC process state for clients that don't need step granularity:

- `InProgress` — a required step is actionable by the user (`NotStarted` / `InProgress` / `Failed` / `Outdated`)
- `PendingReview` — a required step is in backend review (`Finished` / `InternalReview` / `ExternalReview` / `ManualReview` / `PartiallyApproved` / `DataRequested` / `Paused` / `OnHold`)
- `Completed` — all required steps `Completed`
- `Failed` — KYC terminated

Computed in `KycInfoMapper.computeProcessStatus` over the same `getUiSteps()` step list the client already receives.

## How it threads through the mapper

`KycInfoMapper.toDto` computes `requiredStepNames` once and threads it through `KycStepMapper.toStep(step, currentStep, requiredStepNames)` and the new `computeProcessStatus` helper. `toStep` gains a required `requiredStepNames` parameter; its single call site is updated in the same change.

## Backwards compatibility

`isRequired` and `processStatus` are new fields on existing response DTOs — old clients ignore them. No reordering or changes to existing fields/enums.

## Tests

`kyc-info.mapper.spec.ts` (new) covers `computeProcessStatus` via `KycInfoMapper.toDto` — terminated → `Failed`, not-started → `InProgress`, all-completed → `Completed`, in-review → `PendingReview`, `DataRequested` → `PendingReview`, actionable-beats-pending — plus the `isRequired` flag. 10 tests.

One behaviour is pinned explicitly: when a required Ident step has a `Completed` sequence alongside a parallel `Outdated`/`InProgress` sequence, `getUiSteps()`/`sortSteps()` collapses the group to the `Completed` step, so `processStatus` reports `Completed`. The spec asserts and documents this — see the review comment below.

## Local verification

- [x] `npm run type-check` — clean
- [x] `npm run lint` — clean
- [x] `npx jest src/subdomains/generic/kyc/dto/mapper/__tests__/kyc-info.mapper.spec.ts` — 10/10 passing

## Test plan

- [ ] `GET /v2/kyc` — `processStatus` matches the visible state in the admin tool for fixture accounts (completed, mid-review, not-started, terminated).
- [ ] `GET /v2/kyc` — each step's `isRequired` matches `requiredKycSteps` for that account.

## Notes / follow-ups

- `canTrade` was part of an earlier revision of this PR and has been removed: a precomputed trading-permission boolean duplicates the actual trade-endpoint gating (`kycLevel`-based) and can only drift from it — the trade/quote endpoints stay the single authority.
- The Wave-3 fields (`UserV2Dto.capabilities`, `RealUnitRegistrationStatus.ALREADY_REGISTERED`) shipped in #3733.

## Companion app PR

App-side consumption + local-logic removal follows once this merges.


TaprootFreak (commented):
## Review findings

### 1. `canTrade` should be removed — it duplicates endpoint logic and diverges from it
`canTrade` is a precomputed boolean that re-implements the trade-permission decision a second time. The single authority is the API itself: the real trade/quote/buy path gates **solely** on `kycLevel >= 30/50` (`aml-helper.service.ts`, `swap.service.ts`, `buy.controller.ts`, `transaction-helper.ts`, `faucet-request.service.ts`). `computeCanTrade` additionally blocks on Ident/FinancialData step status. Result: the incident user cited in this PR (kycLevel 53, outdated IDENT) is allowed to trade server-side, but `canTrade` reports `false`. A precomputed flag that duplicates endpoint logic can only ever drift — the endpoint is the single source of truth. The client should react to the actual endpoint response instead of pre-checking a boolean.

### 2. `processStatus` — same category, internally inconsistent
`computeProcessStatus` and `computeCanTrade` classify `DATA_REQUESTED` differently (actionable → `IN_PROGRESS` vs. blocking via `isInReview`). Both status sets in `computeProcessStatus` also miss `FINISHED`, `PARTIALLY_APPROVED` and `PAUSED` — a required step in one of those states falls through both `.some()` checks and yields `processStatus = Completed` while the step is still in review.

### 3. `@ApiProperty` doc of `canTrade` is incomplete
The doc lists 3 blocking statuses, but `isInReview` alone expands to FINISHED, EXTERNAL/INTERNAL/MANUAL_REVIEW, PARTIALLY_APPROVED, DATA_REQUESTED, PAUSED — effectively 11+ blocking statuses.

### 4. Merge conflict with #3733
Both PRs modify `user-dto.mapper.ts`, `user-v2.dto.ts` and `user-dto.mapper.spec.ts`. A merge order needs to be fixed and the second PR rebased.

### Minor
The new third `toStep` parameter is justified with "~hundred existing call sites", but there is exactly one caller (`kyc-info.mapper.ts`).

davidleomay (approved):

PR #3733, opened 2026-05-21
feat(user/realunit): expose user capabilities + structured ALREADY_REGISTERED status

Blume1977:
## Summary

Wave 3 of the realunit-app API-as-Decision-Authority plan ([`DFXswiss/realunit-app:docs/api-authority-plan.md`](https://github.com/DFXswiss/realunit-app/blob/develop/docs/api-authority-plan.md)). Two changes that let the realunit-app stop interpreting backend state into UI affordances.

### `UserV2Dto.capabilities`

New `UserCapabilitiesDto` field on the `/v2/user` response. Surfaces per-action flags the realunit-app cubits were re-deriving locally from KYC step status:

- `canEditName` / `canEditAddress`: `false` once `PersonalData` is in any review or completed state (data is locked to keep client and KYC attestation aligned).
- `canEditMail` / `canEditPhone`: `false` only on KYC-terminated accounts.
- `supportAvailable`: requires a verified mail.

`UserDtoMapper.computeCapabilities` mirrors the rules the realunit-app cubits encode today (`settings_user_data_page.dart:239`, `settings_edit_name_cubit.dart:22`, `settings_contact_page.dart:54-67`).

### `RealUnitRegistrationStatus.ALREADY_REGISTERED`

New enum value. Since #3731 the idempotent path (same wallet + same EIP-712 signature) already returns a success status instead of throwing — it returned `COMPLETED`. This PR refines that return value to a dedicated `ALREADY_REGISTERED`, so the realunit-app can distinguish "registration just completed in this call" from "registration was already in place" and skip the post-registration onboarding screens on retry.

This is a change to an existing return value, not a purely additive one: the idempotent completed-retry path now returns `ALREADY_REGISTERED` where it previously returned `COMPLETED`. The HTTP status is unchanged — `completeRegistration` / `completeRegistrationForWalletAddress` map both `COMPLETED` and `ALREADY_REGISTERED` to `201`, so clients that branch on the status code see no change; the distinction is carried by the `status` body field.

The signature-mismatch behaviour from #3731 is preserved — same wallet + different signature still throws `400`; same wallet + same signature is the idempotent `ALREADY_REGISTERED` path.

## What is intentionally NOT in this PR

- **`requiredWorkflow` on `SellPaymentInfoDto`** was considered (audit V9) but BitBox vs software-wallet selection is unavoidably a client-side device fact — the API can't substitute for the client knowing what hardware it has. Will be addressed differently in a future iteration (e.g. by exposing `supportedSignMethods` rather than a single workflow choice).

## Backwards compatibility

- `capabilities` is purely additive — old clients ignore the new field and continue to derive editability locally.
- `ALREADY_REGISTERED` is a new enum value, but on the idempotent completed-retry path it **replaces** the `COMPLETED` value returned since #3731. A client that explicitly checks for `COMPLETED` on that path will now see `ALREADY_REGISTERED`; clients that switch on the `status` field with a fallback treat the unknown value like `FORWARDING_FAILED` (no worse than a generic non-error fallback). The HTTP status stays `201`, so status-code-based clients are unaffected. The companion realunit-app PR consumes the new value.

## Tests

- `user-dto.mapper.spec.ts` adds a `mapUser: capabilities` block covering all five flags across happy path, PersonalData-locked, KYC-terminated, and no-mail fixtures.
- `realunit.service.spec.ts` asserts the idempotent path returns `ALREADY_REGISTERED`.

## Local verification

- [x] `npm run type-check` — clean
- [x] `npm run lint` — clean
- [x] `npm test` — **943 / 943 passing** (938 baseline + 5 new tests)

## Test plan (manual, DEV)

- [ ] `GET /v2/user` for a clean Level-20 user → `capabilities.canEditName=true`.
- [ ] `GET /v2/user` for a user with PersonalData in `ManualReview` → `capabilities.canEditName=false`.
- [ ] `GET /v2/user` for a no-mail user → `capabilities.supportAvailable=false`.
- [ ] `POST /v1/realunit/register/wallet` for an already-registered wallet (same signature) → `201 { "status": "already_registered" }` (returned `201 { "status": "completed" }` before this PR).

## Companion app PR

Will open immediately on the realunit-app side to consume `capabilities` + `ALREADY_REGISTERED`.

### DIFF
--- a/src/subdomains/generic/user/models/user/dto/__tests__/user-dto.mapper.spec.ts
+++ b/src/subdomains/generic/user/models/user/dto/__tests__/user-dto.mapper.spec.ts
@@ -150,9 +150,7 @@ describe('UserDtoMapper', () => {
 
     it('canEditName is false once PersonalData is completed', () => {
       const userData = buildUserData();
-      userData.kycSteps = [
-        buildStep(KycStepName.PERSONAL_DATA, ReviewStatus.COMPLETED),
-      ];
+      userData.kycSteps = [buildStep(KycStepName.PERSONAL_DATA, ReviewStatus.COMPLETED)];
 
       const result = UserDtoMapper.mapUser(userData);
 
@@ -162,9 +160,7 @@ describe('UserDtoMapper', () => {
 
     it('canEditName is false while PersonalData is in review', () => {
       const userData = buildUserData();
-      userData.kycSteps = [
-        buildStep(KycStepName.PERSONAL_DATA, ReviewStatus.MANUAL_REVIEW),
-      ];
+      userData.kycSteps = [buildStep(KycStepName.PERSONAL_DATA, ReviewStatus.MANUAL_REVIEW)];
 
       const result = UserDtoMapper.mapUser(userData);
 
@@ -180,9 +176,7 @@ describe('UserDtoMapper', () => {
     });
 
     it('all edit flags collapse to false on KYC-terminated accounts', () => {
-      const result = UserDtoMapper.mapUser(
-        buildUserData({ kycLevel: KycLevel.REJECTED }),
-      );
+      const result = UserDtoMapper.mapUser(buildUserData({ kycLevel: KycLevel.REJECTED }));
 
       expect(result.capabilities.canEditMail).toBe(false);
       expect(result.capabilities.canEditPhone).toBe(false);
--- a/src/subdomains/generic/user/models/user/dto/__tests__/user-dto.mapper.spec.ts
+++ b/src/subdomains/generic/user/models/user/dto/__tests__/user-dto.mapper.spec.ts
@@ -1,8 +1,14 @@
+import { ConfigService } from 'src/config/config';
 import { Country } from 'src/shared/models/country/country.entity';
+import { createDefaultFiat } from 'src/shared/models/fiat/__mocks__/fiat.entity.mock';
+import { Language } from 'src/shared/models/language/language.entity';
+import { KycStep } from 'src/subdomains/generic/kyc/entities/kyc-step.entity';
+import { KycStepName } from 'src/subdomains/generic/kyc/enums/kyc-step-name.enum';
+import { ReviewStatus } from 'src/subdomains/generic/kyc/enums/review-status.enum';
 import { Organization } from '../../../organization/organization.entity';
 import { AccountType } from '../../../user-data/account-type.enum';
 import { UserData } from '../../../user-data/user-data.entity';
-import { UserDataStatus } from '../../../user-data/user-data.enum';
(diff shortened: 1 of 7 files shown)


TaprootFreak (commented):
## Review findings

### 1. PR description is factually inaccurate
The description states `completeRegistration` / `completeRegistrationForWalletAddress` "return `ALREADY_REGISTERED` instead of throwing `BadRequestException`". No `throw` is removed in this diff. Since #3731 the idempotent path already returns `COMPLETED` via `idempotentRegistrationResult`. This PR only changes that existing return value `COMPLETED → ALREADY_REGISTERED`. It is therefore a behavior change to an existing return value, not a purely additive change — the description should be corrected so reviewers assess it correctly.

### 2. Undocumented HTTP status code change
The controller maps `COMPLETED → 201 CREATED`, everything else → `202 ACCEPTED`. With the idempotent "already registered" path now returning `ALREADY_REGISTERED`, `POST /register/complete` and `/register/wallet` answer **202 instead of 201** in that case. 202 is arguably more correct for an idempotent hit, but it is a wire-visible change for clients that branch on the status code — it should be a conscious, documented decision.

### 3. Merge conflict with #3732
Both PRs modify `user-dto.mapper.ts`, `user-v2.dto.ts` and `user-dto.mapper.spec.ts`.

TaprootFreak (approved):
Approving — both findings from my earlier review are addressed.

**Description** corrected: the `ALREADY_REGISTERED` section no longer frames this as "return instead of throw" (since #3731 the idempotent path already returned a status, not a 400). It now accurately describes the change as `COMPLETED → ALREADY_REGISTERED` on the idempotent retry path, and the backwards-compatibility section reflects that this replaces an existing return value rather than being purely additive.

**HTTP status**: pushed `fix(realunit): return 201 for ALREADY_REGISTERED like COMPLETED` (f3aeeaf8f). `ALREADY_REGISTERED` was falling through to `202 ACCEPTED`; it now maps to `201` like `COMPLETED`, so the idempotent path keeps the `201` it returned before this PR and the `status` body field carries the distinction. type-check / lint / realunit + user-dto specs verified locally; CI green.

The feature code itself is sound — the capabilities mapper and the `ALREADY_REGISTERED` refinement both check out.

Non-blocking note for merge: this branch is ~4 commits behind `develop`, which has since done the MSSQL→PostgreSQL cutover. It auto-merges cleanly and the diff is DB-agnostic, but consider "Update branch" + a green CI run on the merge result before merging. #3732 will also need a rebase against this once it lands (shared `user-dto.mapper.ts` / `user-v2.dto.ts`).

davidleomay (approved):

